PR #1, opened 2017-04-25
[DCOS_OSS-1008] Initial test framework

adamtheturtle:
## High Level Description

This brings a new, experimental testing harness.

## Related Issues

  - [DCOS_OSS-1008](https://jira.mesosphere.com/browse/DCOS_OSS-1008) Build rudimentary DC/OS config-aware integration test framework based on dcos-docker
  
## Future

See README.md for the context of this repository.
I would be happy to put this into a `dcos` owned repository and have asked how to do this.


README.md
```diff
@@ -1,2 +1,79 @@
...
+End to end tests are tests which require a DC/OS cluster to run against.
+Each test spins up at least one cluster, and has the choice of configuring this cluster as appropriate.
+For example, a test may require a cluster with a certain number of agents, or certain configuration options.
+
+The tests should be not be tied to the backend infrastructure.
+That is, they should pass against clusters on all supported infrastructures.
+The current implementation supports only a [DC/OS Docker](https://github.com/dcos/dcos-docker) backend.
+
+Usually tests are kept with the source code.
+However, this is a proof of concept.
```
vespian: `Usually tests are kept with the source code, however, this is just a proof of concept?`

README.md
```diff
@@ -1,2 +1,79 @@
...
+We plan to iterate on this repository and manually run tests.
+With that experience, we will choose where to put the test suite and whether it should be run on CI.
```
vespian: Why do you put every sentence on a new line?

adamtheturtle: The renderer doesn't render the new line characters.
This is purely for maintainability.
The idea comes from http://rhodesmill.org/brandon/2012/one-sentence-per-line/ and I personally have found that it is a useful practice.

The idea is that often a sentence is changed alone.
In normal prose style, the whole paragraph may be deleted and re-added in a `git` diff.
In this style, only the change is visible in the diff.

I have also found that the practice highlights to me, when writing, that a sentence is very long and so maybe it should be split up.

mypy.ini
```diff
@@ -0,0 +1,9 @@
+[mypy]
```
adamtheturtle: Personally I think that any and all checks we can get are highly valuable here. That is because these tests are so slow that finding bugs late is expensive in terms of development time.

README.md
```diff
@@ -1,2 +1,79 @@
...
+
+## Running tests
+
+The tests must be run on a host which is supported by DC/OS Docker.
+One way to guarantee this support is to create a Vagrant VM which supports DC/OS Docker.
+
+```sh
+cd tests/end_to_end/
```
mhrabovcin: This should be updated to reflect that code now lives in a separate repo.

bootstrap.sh
```diff
@@ -0,0 +1,38 @@
...
+	wget \
+	curl \
+	llvm \
+	tar \
+	unzip \
+	git \
+	patch \
+	gcc
+
+curl -L https://raw.githubusercontent.com/pyenv/pyenv-installer/master/bin/pyenv-installer | bash
```
mhrabovcin: I am not sure if its safe to assume that every developer would be ok with having installed software and modified `.bash_profile`, I'd be definitely surprised.

adamtheturtle: This is specifically for the Vagrant VM. I will make that clear in a comment / by renaming the file.

testtools.py
```diff
@@ -0,0 +1,254 @@
...
+            agents=agents,
+            public_agents=public_agents,
+            extra_config=extra_config,
+            generate_config_url=tests_config['dcos_generate_config_url'],
+            generate_config_path=generate_config_path,
+            dcos_docker_path=Path(tests_config['dcos_docker_path']),
+        )
+        self._backend.postflight()
+
+    def __enter__(self) -> 'Cluster':
```
mhrabovcin: Out of curiosity, why is the return type in single quotes?

adamtheturtle: Because this is in the `Cluster` class, `Cluster` is not defined at the time that this is defined. This seems to be the best way to define this type. See _numerous_ discussions online e.g. https://github.com/python/typing/issues/34.

sample-configuration.yaml
```diff
@@ -0,0 +1,3 @@
+dcos_generate_config_url: null
```
adamtheturtle: FWIW a follow-up will get rid of this and make `generate_config_path` required. Tests should not rely on the network where possible. Also, the benefit of just putting a URL here is outweighed by the pain on the second test run. Also the confusion of "what happens when I provide a URL and a path?" is not worth it.

PR #236, opened 2017-08-23
[WIP] Added popen_as_root

rdeliallisi:
Added a way to run commands on nodes asynchronously. This is done by just opening a `PIPE` to the created subprocess without reading or writing to it (which previously caused blocking). 

edit by @adamtheturtle: This is related to https://jira.mesosphere.com/browse/DCOS-18068

### DIFF
--- a/src/dcos_e2e/_common.py
+++ b/src/dcos_e2e/_common.py
@@ -3,6 +3,8 @@ Common utilities for end to end tests.
 """
 
 import logging
+from ipaddress import IPv4Address
+from pathlib import Path
 from subprocess import (
     PIPE,
     STDOUT,
@@ -10,7 +12,7 @@ from subprocess import (
     CompletedProcess,
     Popen,
 )
-from typing import List, Optional, Union
+from typing import Dict, List, Optional, Union
 
 logging.basicConfig(level=logging.DEBUG)
 LOGGER = logging.getLogger(__name__)
@@ -76,3 +78,60 @@ def run_subprocess(
                 retcode, args, output=stdout, stderr=stderr
             )
     return CompletedProcess(args, retcode, stdout, stderr)
+
+
+def compose_ssh_command(
+    ip_address: IPv4Address,
+    ssh_key_path: Path,
+    args: List[str],
+    env: Optional[Dict]=None,
+) -> List[str]:
+    """
+    Run the specified command on the given host using ssh.
+
+    Args:
+        ip_address: The IP address of the node.
+        ssh_key_path: The path to an SSH key which can be used to SSH to
+            the node as the `root` user.
+        args: The command to run on the node.
+        env: Environment variables to be set on the node before running
+                the command. A mapping of environment variable names to
+                values.
+
+    Returns:
+        Full ssh command to be run (ssh arguments + environment variables +
+        other arguments).
+    """
+    env = dict(env or {})
+
+    command = []
+
+    for key, value in env.items():
+        export = "export {key}='{value}'".format(key=key, value=value)
+        command.append(export)
+        command.append('&&')
+
+    command += args
+
+    ssh_args = [
+        'ssh',
+        # Suppress warnings.
+        # In particular, we don't care about remote host identification
+        # changes.
+        '-q',
+        # The node may be an unknown host.
+        '-o',
+        'StrictHostKeyChecking=no',
+        # Use an SSH key which is authorized.
+        '-i',
+        str(ssh_key_path),
+        # Run commands as the root user.
+        '-l',
+        'root',
+        # Bypass password checking.
+        '-o',
+        'PreferredAuthentications=publickey',
+        str(ip_address),
+    ] + command
+
+    return ssh_args
--- a/tests/test_node.py
+++ b/tests/test_node.py
@@ -5,6 +5,7 @@ Tests for managing DC/OS cluster nodes.
 import logging
 from pathlib import Path
 from subprocess import CalledProcessError
+from time import sleep
 
 import pytest
 from pytest_catchlog import CompatLogCaptureFixture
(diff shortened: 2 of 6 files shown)


rdeliallisi:
@adamtheturtle The new test passes. What fails are 2, most likely unrelated, tests.

adamtheturtle:
The tests which have failed are the tests which use the DC/OS Enterprise artifact.
The DC/OS Artifact URL is private, and so it is encrypted as a secret on the repository.
As your PR comes from a fork, it does not have access to the secrets [for security reasons](https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions).
Therefore, what I will do it merge this not into `master` but into a different branch of DC/OS E2E, and then the tests will run.

tests/test_node.py
```diff
@@ -71,3 +72,46 @@ def test_run_as_root(
...
+            generate_config_path=oss_artifact,
+        ) as cluster:
+            (master, ) = cluster.masters
+            popen_1 = master.popen_as_root(
+                args=[
+                    '(mkfifo', '/tmp/pipe', '|', 'true)'
+                    '&&', '(cat', '/tmp/pipe)'
+                ]
+            )
+            sleep(1)
```
adamtheturtle: I don't understand why there needs to be a sleep - can this be removed? Or a comment added.

adamtheturtle:
@rdeliallisi I have added you as a collaborator on the upstream repository so you don't have this test issue.

src/dcos_e2e/_common.py
```diff
@@ -76,3 +78,60 @@ def run_subprocess(
                 retcode, args, output=stdout, stderr=stderr
             )
     return CompletedProcess(args, retcode, stdout, stderr)
+
+
+def compose_ssh_command(
```
adamtheturtle: This `_common` module is used for functions and classes which are needed by multiple modules - they are "common" or "shared" between modules. This utility is used only in the `node` module, and so it should be a private function or method there.

src/dcos_e2e/_common.py
```diff
@@ -76,3 +78,60 @@ def run_subprocess(
...
+    ip_address: IPv4Address,
+    ssh_key_path: Path,
+    args: List[str],
+    env: Optional[Dict]=None,
+) -> List[str]:
+    """
+    Run the specified command on the given host using ssh.
```
adamtheturtle: `s/ssh/SSH`.

src/dcos_e2e/_common.py
```diff
@@ -76,3 +78,60 @@ def run_subprocess(
...
+        ip_address: The IP address of the node.
+        ssh_key_path: The path to an SSH key which can be used to SSH to
+            the node as the `root` user.
+        args: The command to run on the node.
+        env: Environment variables to be set on the node before running
+                the command. A mapping of environment variable names to
+                values.
+
+    Returns:
+        Full ssh command to be run (ssh arguments + environment variables +
```
adamtheturtle: `s/ssh/SSH`.

tests/test_node.py
```diff
@@ -71,3 +72,46 @@ def test_run_as_root(
...
+        with Cluster(
+            agents=0,
+            public_agents=0,
+            cluster_backend=cluster_backend,
+            generate_config_path=oss_artifact,
+        ) as cluster:
+            (master, ) = cluster.masters
+            popen_1 = master.popen_as_root(
+                args=[
+                    '(mkfifo', '/tmp/pipe', '|', 'true)'
```
adamtheturtle: I don't really understand this and I also am not sure why the original test we paired on does not work.

The test we wrote:

`popen_1: Loop until FOO exists; cat FOO`
`popen_2: Create FOO with content BAR`

Assert that `popen_1` and `popen_2` finish and `popen_1` includes the output `BAR`.

I'm happy to discuss this on Slack or voice chat if it is simpler.

tests/test_node.py
```diff
@@ -71,3 +72,46 @@ def test_run_as_root(
...
+    # An arbitrary time slice that is longer than it takes to spin up a cluster
+    # but avoids bash infinite loops.
+    @pytest.mark.timeout(60 * 15)
+    def test_popen_as_root(
+        self,
+        cluster_backend: ClusterBackend,
+        oss_artifact: Path,
+    ) -> None:
+        """
+        It is possible to run commands as root asynchronously
```
adamtheturtle: `s/asynchronously/asynchronously.`

adamtheturtle:
Thanks @rdeliallisi , this is great and will be very useful. Please address the comments 😄

tests/test_node.py
rdeliallisi: No need for it, the entire magic in the commands has the sole purpose to remove this sleep command and make test deterministic.

tests/test_node.py
rdeliallisi: The problem with the original test is that, `while` is not a construct available on all shells and thus our command would not run at all in a python subprocess. This could be remedied by setting `shell=True` for the `Popen` command. This lead to the next problem. The command we pass to `popen_as_root` via `args` is run via `SSH` which did not support running while loops, so I just dropped the idea.

Furthermore I think this idea is cleaner. It uses `named pipes` which are blocking mechanisms. If a process tries to read from a pipe it blocs until there is a process writing on it and vice versa, if a process writes to a pipe, it blocks until there is a process reading from it. In our case we have:
`popen_1: Create a pipe if not existing and read from it.`
`popen_2: Create a pipe if not existing and write to it.`
The commands could have been much simpler but we have to take care of situations where the processes do not start in the specified order.

rdeliallisi:
Work continues on https://github.com/adamtheturtle/dcos-e2e/pull/237. Closing this PR.

PR #237, opened 2017-08-24
Added popen_as_root

rdeliallisi:
Added a way to run commands on nodes asynchronously. This is done by just opening a PIPE to the created subprocess without reading or writing to it (which previously caused blocking). This is related to https://jira.mesosphere.com/browse/DCOS-18068.


rdeliallisi:
@adamtheturtle Any other suggestions?

adamtheturtle:
@rdeliallisi Thank you, I'm really happy with this change.

For the record, I have made sure that a comment on the JIRA issue links to this PR.

What remains is documentation in https://github.com/adamtheturtle/dcos-e2e/blob/master/API.md#methods-1. I did not think of that before, and it is the only remaining work that I can think of.

gpaul:
Very cool

PR #238, opened 2017-08-28
DC/OS E2E - Readme fixes

rdeliallisi:
Fix `API.md` and add pointers to `CONTRIBUTING.md` in `README.md` for unit tests.

### DIFF
--- a/API.md
+++ b/API.md
@@ -243,7 +243,3 @@ The method returns a `Popen` object that can be used to communicate to the under
 #### `ip_address`
 
 The IP address of the node.
-
-#### `ssh_key_path`
-
-The path to an SSH key which can be used to SSH to the node as the `root` user.
--- a/API.md
+++ b/API.md
@@ -243,3 +243,7 @@ The method returns a `Popen` object that can be used to communicate to the under
 #### `ip_address`
 
 The IP address of the node.
+
+#### `ssh_key_path`
+
+The path to an SSH key which can be used to SSH to the node as the `root` user.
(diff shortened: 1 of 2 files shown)


API.md
```diff
@@ -243,3 +243,7 @@ The method returns a `Popen` object that can be used to communicate to the under
 #### `ip_address`
 
 The IP address of the node.
+
+#### `ssh_key_path`
```
adamtheturtle: This documentation is for the public interface of a node. There is no `ssh_key_path` attribute. There is only a `_ ssh_key_path` and private interfaces are not documented.

adamtheturtle: That is - the preferred action is to remove this block.

README.md
```diff
@@ -79,7 +79,8 @@ See [`CONTRIBUTING.md`](./CONTRIBUTING.md) for details on how to contribute to t
 ## Test Environment
 
 Tests for this package and tests which use this package must be run on a host which is supported by DC/OS Docker.
-See the [DC/OS Docker README](https://github.com/dcos/dcos-docker/blob/master/README.md).
+For more information about `DC/OS Docker` see [DC/OS Docker README](https://github.com/dcos/dcos-docker/blob/master/README.md).
```
adamtheturtle: `s/see/see the`.

API.md
rdeliallisi: Ok, will remove.

adamtheturtle: Also, FWIW, this would fail CI because of the table of contents. Unfortunately there is currently no way for `make lint` to verify that the ToC is correct locally, only on Travis (see the `Makefile` for details). I think one has to `make lint` after modifying any headings.

adamtheturtle:
Thanks @rdeliallisi - please make the minor changes and then merge after the lint phase is complete on one builder.

PR #254, opened 2017-09-21
Volume and SSH fixes

rdeliallisi:
* Mount volume needed for `files_to_copy_to_masters` in `rw` mode.
* Add `-o IdentitiesOnly=yes` to the SSH command. This makes sure that the keys already present in the keychain are ignored and only the ones passed with the `-i` option are used.

### DIFF
--- a/src/dcos_e2e/backends/_dcos_docker/__init__.py
+++ b/src/dcos_e2e/backends/_dcos_docker/__init__.py
@@ -183,6 +183,8 @@ class DCOS_Docker_Cluster(ClusterManager):  # pylint: disable=invalid-name
 
         master_mounts = []
         for host_path, master_path in files_to_copy_to_masters.items():
+            # The volume is mounter `rw` because certain processes change the
+            # content or permission of the files on the volume.
             mount = '-v {host_path}:{master_path}:rw'.format(
                 host_path=host_path.absolute(),
                 master_path=master_path,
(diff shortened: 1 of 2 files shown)


src/dcos_e2e/backends/_dcos_docker/__init__.py
```diff
@@ -183,7 +183,7 @@ def __init__(  # pylint: disable=super-init-not-called,too-many-statements
 
         master_mounts = []
         for host_path, master_path in files_to_copy_to_masters.items():
-            mount = '-v {host_path}:{master_path}:ro'.format(
+            mount = '-v {host_path}:{master_path}:rw'.format(
```
adamtheturtle: ❓ Optionally add a comment about why rw was chosen / required.

src/dcos_e2e/node.py
```diff
@@ -63,6 +63,9 @@ def compose_ssh_command(
             # In particular, we don't care about remote host identification
             # changes.
             '-q',
+            # This makes sure that only keys passed wih the -i option are used.
```
adamtheturtle: Typo `s/wih/with`.
Can you please expand the comment to say why we want this.

adamtheturtle:
Thanks @rdeliallisi - please address the comments and I should merge once the tests pass.

rdeliallisi:
@adamtheturtle Comments fixed. Waiting for CI to go green.

adamtheturtle:
@rdeliallisi The typo "wih" (instead of "with") persists.

PR #256, opened 2017-10-04
Bump version

rdeliallisi:


CHANGELOG.md
```diff
@@ -22,6 +23,12 @@
 
 # Changelog
 
+## 2017.10.04.0
+
+* Added Apache2 license.
+* Repository moved to /mesosphere/dcos-e2e.
```
adamtheturtle: Put the full URL as this is not guaranteed to be GitHub.

CHANGELOG.md
```diff
@@ -22,6 +23,12 @@
...
+* Add 2 new functions - run and popen, which have customisable users.
```
adamtheturtle: * Added `run`, which is similar to `run_as_root` but takes a `user` argument.
* Added `popen`, which can be used for running commands asynchronously.

PR #292, opened 2017-10-25
Typo fix

bamarni:


adamtheturtle:
Thank you Bilal. I will merge this despite https://mesosphere.slack.com/archives/C07D97YFL/p1509015585000058 and accept risk. However, this project is not amenable to forks. That is because many of the tests use Travis secrets. I can add you as a contributor at your request and then you can create and push branches to the repository directly.

PR #307, opened 2017-11-06
[WIP] Replaced generate_config_path with url

timaa2k:
In order to support the DC/OS advanced installation method which requires the build artifact to reside on a HTTP(S) server we need to allow for artifacts being passed to DC/OS E2E by url.
The documentation of any cluster backend must specify which url schemes it will accept.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,27 +14,30 @@ env:
   # per commit.
   # We run almost one builder per test.
   matrix:
-      - "TEST_PATTERN=tests/test_enterprise.py::TestEnterpriseIntegrationTests::test_run_pytest"
-      - "TEST_PATTERN=tests/test_enterprise.py::TestWaitForDCOS::test_auth_with_cli"
-      - "TEST_PATTERN=tests/test_enterprise.py::TestCopyFiles::test_copy_files_to_installer"
+      # - "TEST_PATTERN=tests/test_enterprise.py::TestEnterpriseIntegrationTests::test_run_pytest"
+      # - "TEST_PATTERN=tests/test_enterprise.py::TestWaitForDCOS::test_auth_with_cli"
+      # - "TEST_PATTERN=tests/test_enterprise.py::TestCopyFiles::test_copy_files_to_installer"
       # This test class uses one scoped cluster so we run it on one builder.
-      - "TEST_PATTERN=tests/test_node.py::TestNode"
-      - "TEST_PATTERN=tests/test_harness.py::TestIntegrationTests::test_run_pytest"
-      - "TEST_PATTERN=tests/test_harness.py::TestExtendConfig::test_extend_config"
-      - "TEST_PATTERN=tests/test_harness.py::TestExtendConfig::test_default_config"
-      - "TEST_PATTERN=tests/test_harness.py::TestClusterSize::test_default"
-      - "TEST_PATTERN=tests/test_harness.py::TestClusterSize::test_custom"
-      - "TEST_PATTERN=tests/test_harness.py::TestClusterLogging::test_live_logging"
-      - "TEST_PATTERN=tests/test_harness.py::TestClusterLogging::test_no_live_logging"
-      - "TEST_PATTERN=tests/test_harness.py::TestMultipleClusters::test_two_clusters"
-      - "TEST_PATTERN=tests/test_harness.py::TestDestroyOnError::test_default_exception_raised"
-      - "TEST_PATTERN=tests/test_harness.py::TestDestroyOnError::test_set_false_exception_raised"
-      - "TEST_PATTERN=tests/test_harness.py::TestDestroyOnSuccess::test_default"
-      - "TEST_PATTERN=tests/test_harness.py::TestDestroyOnSuccess::test_false"
-      - "TEST_PATTERN=tests/backends/test_docker.py::TestCustomMasterMounts::test_custom_master_mounts"
-      - "TEST_PATTERN=tests/backends/test_existing_cluster.py::TestExistingCluster::test_existing_cluster"
+      # - "TEST_PATTERN=tests/test_node.py::TestNode"
+      # - "TEST_PATTERN=tests/test_harness.py::TestIntegrationTests::test_run_pytest"
+      # - "TEST_PATTERN=tests/test_harness.py::TestExtendConfig::test_extend_config"
+      # - "TEST_PATTERN=tests/test_harness.py::TestExtendConfig::test_default_config"
+      # - "TEST_PATTERN=tests/test_harness.py::TestClusterSize::test_default"
+      # - "TEST_PATTERN=tests/test_harness.py::TestClusterSize::test_custom"
+      # - "TEST_PATTERN=tests/test_harness.py::TestClusterLogging::test_live_logging"
+      # - "TEST_PATTERN=tests/test_harness.py::TestClusterLogging::test_no_live_logging"
+      # - "TEST_PATTERN=tests/test_harness.py::TestMultipleClusters::test_two_c
+      # lusters"
+      # - "TEST_PATTERN=tests/test_harness.py::TestDestroyOnError::test_default_exception_raised"
+      # - "TEST_PATTERN=tests/test_harness.py::TestDestroyOnError::test_set_false_exception_raised"
+      # - "TEST_PATTERN=tests/test_harness.py::TestDestroyOnSuccess::test_default"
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -52,7 +51,7 @@ class TestExample:
         with Cluster(
             extra_config={'check_time': True},
             cluster_backend=Docker(),
-            generate_config_path=Path('/tmp/dcos_generate_config.sh'),
+            generate_config_url=Path('file:///tmp/dcos_generate_config.sh'),
```
adamtheturtle: Remove `Path`

tests/backends/test_docker.py
```diff
@@ -54,3 +56,73 @@ def test_custom_master_mounts(
...
+
+
+class TestBadParameters:
+    """
+    Tests for bad parameters passed to Docker clusters.
+    """
+
+    def test_no_build_artifact(self, tmpdir: local) -> None:
+        """
+        The docker backend requires a build artifact in order
```
adamtheturtle: `s/docker/Docker/g`

tests/backends/test_docker.py
```diff
@@ -54,3 +56,73 @@ def test_custom_master_mounts(
...
+                cluster_backend=Docker(workspace_dir=tmpdir),
+                build_artifact=None,
+                masters=1,
+                agents=0,
+                public_agents=0,
+            ):
+                pass  # pragma: no cover
+
+        expected_error = (
+            'The Docker backend only supports creating new clusters. '
```
adamtheturtle: Instead, make it really clear what a user should do. Perhaps "The Docker backend only supports creating new clusters. `build_artifact` must be a path to a build artifact."

tests/backends/test_docker.py
```diff
@@ -54,3 +56,73 @@ def test_custom_master_mounts(
...
+            'receive a build artifact.'
+        )
+
+        assert str(excinfo.value) == expected_error
+
+    def test_unsupported_build_artifact(
+        self, tmpdir: local, oss_artifact_url: str
+    ) -> None:
+        """
+        If the given build artifact is not a Path
```
adamtheturtle: ``s/Path/`Path`/`

tests/backends/test_docker.py
```diff
@@ -54,3 +56,73 @@ def test_custom_master_mounts(
...
+        the Docker backend will not accept it.
```
adamtheturtle: ``s/not accept it/raise a `NotImplementedError```

tests/backends/test_docker.py
```diff
@@ -54,3 +56,73 @@ def test_custom_master_mounts(
...
+
+        expected_error = (
+            'The Docker backend only supports creating new clusters. '
+            'Therefore any cluster created with the backend must '
+            'receive a build artifact.'
+        )
+
+        assert str(excinfo.value) == expected_error
+
+    def test_unsupported_build_artifact(
```
adamtheturtle: It isn't clear what an "unsupported build artifact" is. Instead perhaps `test_build_artifact_does_not_exist`.

There are no lines of code uncovered without this test, right? I'm not sure that it is needed.

tests/backends/test_docker.py
```diff
@@ -54,3 +56,73 @@ def test_custom_master_mounts(
...
+        )
+
+        assert str(excinfo.value) == expected_error
+
+    def test_invalid_build_artifact(self, tmpdir: local) -> None:
+        """
+        If the given Path does not point to a valid build artifact
+        the subprocess for calling DC/OS Docker will fail.
+        """
+        invalid_build_artifact = Path('/{}'.format(uuid.uuid4()))
```
adamtheturtle: The Mesosphere style guide suggests that we use named format arguments. E.g. `'/{directory_name}'.format(directory_name=uuid.uuid4()`.

tests/backends/test_docker.py
```diff
@@ -54,3 +56,73 @@ def test_custom_master_mounts(
...
+        expected_error = (
+            'The Docker backend only supports creating clusters from '
+            'build artifacts specified by Path.'
+        )
+
+        assert str(excinfo.value) == expected_error
+
+    def test_invalid_build_artifact(self, tmpdir: local) -> None:
+        """
+        If the given Path does not point to a valid build artifact
```
adamtheturtle: In particular this test points to a file that doesn't exist. This isn't new functionality and I think that the test is not necessary. In fact I just removed a very similar test! https://github.com/mesosphere/dcos-e2e/pull/296/files#diff-824a34d44e93c1012e3738b9f4ec110dL21.

src/dcos_e2e/backends/_docker/__init__.py
```diff
@@ -131,7 +131,28 @@ def __init__(  # pylint: disable=super-init-not-called,too-many-statements
         Raises:
             CalledProcessError: The step to create and install containers
                 exited with a non-zero code.
+            ValueError: Raised if `build_artifact`is `None`. The Docker backend
+                requires a path to valid `build_artifact` to install DC/OS on
+                a newly created Docker cluster.
+            NotImplementedError: Raised if `build_artifact` is a url string
+                instead of a Path because the Docker backend only supports
```
adamtheturtle: `` s/Path/`Path` ``.

`s/url/URL`.

BACKENDS.md
```diff
@@ -42,6 +42,10 @@ This is equivalent to `dir` in [TemporaryDirectory](https://docs.python.org/3/li
 Mounts to add to master node containers.
 See `volumes` in [the `docker-py` documentation](http://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.run) for details.
 
+
+When creating a `Cluster` with this backend, the following parameter conditions must be true:
+* `build_artifact` must be of type Path
```
adamtheturtle: `` s/Path/`pathlib.Path` ``

CHANGELOG.md
```diff
@@ -28,6 +28,9 @@
 ## Next
 
 * Backwards incompatible change: Rename `DCOS_Docker` backend to `Docker` backend.
+* Replace `generate_config_path` with `build_artifact` that can either be a
```
adamtheturtle: As above, note that this is a "Backwards incompatible change".

Note that when we do a release to use these changes, we'll have to change existing tests.

adamtheturtle:
Awesome work @timaa2k - as you can see, many of the changes are cosmetic. Let's merge ASAP though so you can continue your work - we don't need to perfect this as it is already a bit of a hack until we have a separate install method.

From our discussion and further thought, I'm thinking that a `ClusterManager` should probably include `install_dcos_from_path` and `install_dcos_from_url` options. We can then `raise NotImplementedError`s where appropriate. That will simplify the tests greatly.

.travis.yml
```diff
@@ -32,6 +32,10 @@ env:
       - "TEST_PATTERN=tests/test_harness.py::TestDestroyOnSuccess::test_default"
       - "TEST_PATTERN=tests/test_harness.py::TestDestroyOnSuccess::test_false"
       - "TEST_PATTERN=tests/backends/test_docker.py::TestCustomMasterMounts::test_custom_master_mounts"
+
+      - "TEST_PATTERN=tests/backends/test_docker.py::TestBadParameters::test_no_build_artifact"
```
adamtheturtle: Each builder takes about 5 minutes and we have a limited number of builders available. I'd copy the format for the other `TestBadParameters` below and put these all on one builder, with a note.

API.md
```diff
@@ -69,9 +69,9 @@ This is a context manager which spins up a cluster.
 The backend to use for the cluster.
 See [`BACKENDS.md`](./BACKENDS.md) for details.
 
-#### `generate_config_path`
+#### `build_artifact`
 
-The path to a build artifact to install.
+The url or path to a build artifact to install.
```
adamtheturtle: `s/url/URL`.

src/dcos_e2e/backends/_base_classes.py
```diff
@@ -30,7 +30,8 @@ def __init__(
         Create a DC/OS cluster with the given `cluster_backend`.
 
         Args:
-            generate_config_path: The path to a build artifact to install.
+            build_artifact: The `Path` or URL string to a build artifact
+            of DC/OS to install from.
```
adamtheturtle: This line should be indented four spaces.

adamtheturtle:
Thanks @timaa2k . Merging.

Please take care of the indentation in a follow-up.

PR #310, opened 2017-11-09
[WIP] default ssh user

timaa2k:
In order to run commands on cluster nodes that do not have root access over SSH enabled by default a cluster advertises the default SSH user that can be used instead.

### DIFF
--- a/API.md
+++ b/API.md
@@ -26,13 +26,13 @@
     - [`masters`](#masters-1)
     - [`agents`](#agents-1)
     - [`public_agents`](#public_agents-1)
+    - [`default_ssh_user`](#default_ssh_user)
 - [`dcos_e2e.node.Node`](#dcos_e2enodenode)
   - [Parameters](#parameters-1)
     - [`ip_address`](#ip_address)
     - [`ssh_key_path`](#ssh_key_path)
   - [Methods](#methods-1)
     - [`node.run(args, user, log_output_live=False, env=None)`](#noderunargs-user-log_output_livefalse-envnone)
-    - [`node.run_as_root(args, log_output_live=False, env=None)`](#noderun_as_rootargs-log_output_livefalse-envnone)
     - [`node.popen(args, user, env=None) -> Popen`](#nodepopenargs-user-envnone---popen)
     - [`node.send_file(local_path, remote_path) -> None`](#nodesend_filelocal_path-remote_path---none)
   - [Attributes](#attributes-1)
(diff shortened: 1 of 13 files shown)


src/dcos_e2e/backends/_docker/__init__.py
```diff
@@ -462,7 +469,7 @@ def __init__(  # pylint: disable=super-init-not-called,too-many-statements
             'ssh_port':
             22,
             'ssh_user':
-            'root',
+            cluster_backend.default_ssh_user,
```
adamtheturtle: I love this because it makes it easier for us to move this config creation into `Cluster`.

API.md
```diff
@@ -26,13 +26,13 @@
     - [`masters`](#masters-1)
     - [`agents`](#agents-1)
     - [`public_agents`](#public_agents-1)
+    - [`default_ssh_user`](#default_ssh_user)
```
adamtheturtle: This is a public change so please add something to the `CHANGELOG`.

src/dcos_e2e/cluster.py
```diff
@@ -116,9 +117,14 @@ def wait_for_dcos(self) -> None:
             '--diag',
         ]
 
+        # Must be run privileged
```
adamtheturtle: I don't like this change, it seems hidden but I'm happy to discuss. Why not just make the user use `sudo` where appropriate?

src/dcos_e2e/cluster.py
```diff
@@ -116,9 +117,14 @@ def wait_for_dcos(self) -> None:
...
+        if not self.default_ssh_user == 'root':
+            diagnostics_args = ['sudo'] + diagnostics_args
```
adamtheturtle: At least for now this code isn't covered by tests, so IMO remove it and we can discuss it for the follow-up.

src/dcos_e2e/node.py
```diff
@@ -118,37 +118,6 @@ def run(
         ssh_args = self._compose_ssh_command(args=args, user=user, env=env)
         return run_subprocess(args=ssh_args, log_output_live=log_output_live)
 
-    def run_as_root(
```
adamtheturtle: Please mention this in the changelog.

API.md
```diff
@@ -169,7 +173,7 @@ The IP address of the node.
 
 #### `ssh_key_path`
 
-The path to an SSH key which can be used to SSH to the node as the `root` user.
+The path to an SSH key which can be used to SSH to the node as the clusters `default_ssh_user` user.
```
adamtheturtle: `s/clusters/cluster's`.

adamtheturtle:
Really lovely work Tim! Just the minor comments to go.

PR #311, opened 2017-11-10
[WIP] Install from path method for the Docker backend

timaa2k:
In order to support test that require modifying files on the cluster nodes before installing DC/OS we would like to have a user initiated installation method. This PR prepares the Docker backend for this functionality.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -96,9 +96,6 @@ disable=line-too-long,
         # This cannot be overridden when pylint is overzealous.
         # See https://github.com/PyCQA/pylint/issues/214.
         duplicate-code,
-        # Disable temporary until split _docker/__init__.py
-	too-many-branches,
-        too-many-instance-attributes,
 
 [REPORTS]
 
--- a/pylintrc
+++ b/pylintrc
@@ -96,7 +96,9 @@ disable=line-too-long,
         # This cannot be overridden when pylint is overzealous.
         # See https://github.com/PyCQA/pylint/issues/214.
         duplicate-code,
+        # Disable temporary until split _docker/__init__.py
 	too-many-branches,
+        too-many-instance-attributes,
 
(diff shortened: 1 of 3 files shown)


pylintrc
```diff
@@ -96,7 +96,9 @@ disable=line-too-long,
         # This cannot be overridden when pylint is overzealous.
         # See https://github.com/PyCQA/pylint/issues/214.
         duplicate-code,
+        # Disable temporary until split _docker/__init__.py
```
adamtheturtle: Instead, put this as a `pylint` specific comment on that one method.

src/dcos_e2e/backends/_docker/__init__.py
```diff
@@ -445,17 +449,42 @@ def __init__(  # pylint: disable=super-init-not-called,too-many-statements
                 tmpfs=node_tmpfs_mounts,
             )
 
+        for node in {*self.masters, *self.agents, *self.public_agents}:
```
adamtheturtle: Note for later: I think that we have no test which tests that this works.

src/dcos_e2e/backends/_docker/__init__.py
```diff
@@ -445,17 +449,42 @@ def __init__(  # pylint: disable=super-init-not-called,too-many-statements
...
+            # Remove stray file that prevents non-root SSH.
+            # https://ubuntuforums.org/showthread.php?t=2327330
+            node.run(
+                args=['rm', '-f', '/run/nologin'],
+                user=cluster_backend.default_ssh_user
+            )
+
+        # Logically a SSH user should be part of a `Node`
```
adamtheturtle: Full stop at end of sentence.

adamtheturtle:
Great, thanks! Please take care of the comments in a follow-up.

PR #313, opened 2017-11-13
[WIP] Split DC/OS installation from Cluster creation

timaa2k:
Splitting the DC/OS installation from the cluster creation allows for a simplified interface of the Cluster class. The change includes explicit installation methods for clusters created with a particular backend. It enables simplification of tests that require to interact with a cluster before installing DC/OS.


README.md
```diff
@@ -50,10 +50,12 @@ class TestExample:
 
     def test_example(self):
         with Cluster(
-            extra_config={'check_time': True},
             cluster_backend=Docker(),
-            build_artifact=Path('/tmp/dcos_generate_config.sh'),
         ) as cluster:
+            cluster.install_dcos_from_path(
```
adamtheturtle: ❤️

src/dcos_e2e/backends/_base_classes.py
```diff
@@ -48,6 +38,28 @@ def __init__(
...
 
     @abc.abstractmethod
+    def install_dcos_from_url(
+        self,
+        build_artifact: str,
+        extra_config: Dict[str, Any],
+        log_output_live: bool,
+    ) -> None:
+        """
+        Stub for the DC/OS advanced installation method.
```
adamtheturtle: Instead of this, give the docstring for what the method should do. For example, `destroy` has:

```
Destroy all nodes in the cluster.
```

tests/test_harness.py
```diff
@@ -265,15 +265,10 @@ def test_two_clusters(
         This is because Travis CI has a space limit which is exceeded if we
```
adamtheturtle: I think that this existing comment is incorrect.

tests/test_enterprise.py
```diff
@@ -30,24 +30,25 @@ def test_run_pytest(
...
 
-        with Cluster(
-            build_artifact=enterprise_artifact,
-            cluster_backend=cluster_backend,
-            extra_config=extra_config,
-            log_output_live=True,
-        ) as cluster:
+        with Cluster(cluster_backend=cluster_backend) as cluster:
+            cluster.install_dcos_from_path(
+                enterprise_artifact,
```
adamtheturtle: Please use named arguments for all arguments here.

tests/test_harness.py
```diff
@@ -233,17 +235,15 @@ def test_no_live_logging(
...
-            # It is not possible to create a cluster with two master nodes.
             with Cluster(
                 masters=2,
                 cluster_backend=cluster_backend,
-                build_artifact=oss_artifact,
-            ):
-                pass  # pragma: no cover
+            ) as cluster:
+                # It is not possible to install DC/OS with two master nodes.
+                cluster.install_dcos_from_path(oss_artifact)
```
adamtheturtle: Instead, put only this line in the `pytest.raises` context manager.

The advantage to this is that it is clearer what raised the error.

tests/test_harness.py
```diff
@@ -212,17 +213,18 @@ def test_live_logging(
         oss_artifact: Path,
     ) -> None:
         """
-        If `log_output_live` is given as `True`, subprocess output is logged.
+        If `log_output_live` is given as `True`, the installation output is
```
adamtheturtle: I think that this test is now in the wrong place. This is:

```
class TestClusterLogging:
    """
    Tests for logs created by the ``Cluster``.
    """
```

Perhaps this should be in a `TestInstallDcosFromPath` class?

CHANGELOG.md
```diff
@@ -33,6 +33,12 @@ that can either be a `Path` or a HTTP(S) URL string. This allows for supporting
 methods that require build artifacts to be downloaded from a HTTP server.
 * Backwards incompatible change: Remove `run_as_root`. Instead require a `default_ssh_user`
 for backends to `run` commands over SSH on any cluster `Node` created with this backend.
+* Backwards incompatible change: Split the DC/OS installation from the ClusterManager
+`__init__` procedure. This allows for installing DC/OS after `Cluster` creation,
+and therefore enables decoupling of transfering files ahead of the installation process.
+* Backwards incompatible change: Explicit distinction of installtion methods by providing
```
adamtheturtle: There's also the backwards incompatible change of not having `log_output_live` on `Cluster` any more.

src/dcos_e2e/backends/_docker/__init__.py
```diff
@@ -457,20 +422,59 @@ def __init__(  # pylint: disable=super-init-not-called,too-many-statements
...
+        self,
+        build_artifact: str,
+        extra_config: Dict[str, Any],
+        log_output_live: bool,
+    ) -> None:
+        """
+        Args:
+            build_artifact: The URL string to a build artifact to install DC/OS
+                from.
+            extra_config: Implementations may come with a "base"
```
adamtheturtle: This is the generic explanation but should instead be specific to this implementation.

src/dcos_e2e/backends/_docker/__init__.py
```diff
@@ -457,20 +422,59 @@ def __init__(  # pylint: disable=super-init-not-called,too-many-statements
...
     ) -> None:
+        """
+        Args:
+            build_artifact: The `Path` to a build artifact to install DC/OS
+                from.
+            extra_config: Implementations may come with a "base"
+                configuration. This dictionary can contain extra installation
+                configuration variables.
+            log_output_live: If `True`, log output of the installation live.
+                If `True`, stderr is merged into stdout in the return value.
```
adamtheturtle: There is no return value, so this isn't correct.

src/dcos_e2e/backends/_docker/__init__.py
```diff
@@ -457,20 +422,59 @@ def __init__(  # pylint: disable=super-init-not-called,too-many-statements
...
+    ) -> None:
+        """
+        Args:
+            build_artifact: The URL string to a build artifact to install DC/OS
+                from.
+            extra_config: Implementations may come with a "base"
+                configuration. This dictionary can contain extra installation
+                configuration variables.
+            log_output_live: If `True`, log output of the installation live.
+                If `True`, stderr is merged into stdout in the return value.
```
adamtheturtle: There is no return value, so this isn't correct.

src/dcos_e2e/cluster.py
```diff
@@ -185,10 +182,65 @@ def default_ssh_user(self) -> str:
...
+            log_output_live: If `True`, log output of the installation live.
+                If `True`, stderr is merged into stdout in the return value.
+
+        Raises:
+            NotImplementedError: `NotImplementedError` because the given
+                backend provides a more efficient installation method than
+                the DC/OS advanced installation method.
+        """
+        self._cluster.install_dcos_from_url(
+            build_artifact,
```
adamtheturtle: Please use named arguments where possible. This is a personal preference of mine and not necessarily a best practice 😄

src/dcos_e2e/backends/_docker/__init__.py
```diff
@@ -457,20 +422,59 @@ def __init__(  # pylint: disable=super-init-not-called,too-many-statements
...
+        Raises:
+            NotImplementedError: `NotImplementedError` because the Docker
+                backend does not support the DC/OS advanced installation
+                method.
+        """
+        message = (
+            'The Docker backend does not support the installation of DC/OS '
+            'by build artifacts passed via URL string. This is because a more '
+            'efficient installation method exists that uses a local build '
+            'artifact.'
```
adamtheturtle: Error messages, where possible, should help the user make the right choice. In this case `install_dcos_from_path` is the right choice, so point users to this.

src/dcos_e2e/backends/_existing_cluster.py
```diff
@@ -101,13 +91,6 @@ def __init__(  # pylint: disable=super-init-not-called
...
-        if build_artifact is not None:
-            message = (
-                'Cluster already exists with DC/OS installed. '
-                'Therefore, `build_artifact` must be `None`.'
-            )
-            raise ValueError(message)
-
         if masters != len(self._masters):
             message = (
                 'The number of master nodes is `1`. '
```
adamtheturtle: Existing bug! This is hardcoded to 1 but really should be a variable.

src/dcos_e2e/backends/_existing_cluster.py
```diff
@@ -144,6 +120,46 @@ def __init__(  # pylint: disable=super-init-not-called
...
+        """
+        Raises:
+            NotImplementedError: Raises `NotImplementedError` because it
+                is assumed that clusters created with the ExistingCluster
+                backend already have an installed instance of DC/OS
+                running on them.
+        """
+        message = (
+            'The ExistingCluster backend does not support the installing '
+            'DC/OS because it is assumed that an instance of DC/OS is '
```
adamtheturtle: ??? I think that "because" is not appropriate here because you could install DC/OS on a cluster already with DC/OS, right?

API.md
```diff
@@ -110,7 +98,45 @@ If set to `True`, the cluster is destroyed on exit if there is no exception rais
...
+
+The HTTP(S) URL string to a build artifact to install.
+
+##### `extra_config`
+
+Configuration variables to add to a base configuration.
+
+##### `log_output_live`
+
+If set to `True`, the output of the DC/OS installation process will be logged.
```
adamtheturtle: Is this definitely right? I thought that it was logged either way but `log_output_live=True` shows that output as it happens, rather than after it has all happened. Please re-add "To see these logs in `pytest` tests, use the `-s` flag.".

BACKENDS.md
```diff
@@ -74,19 +76,21 @@ See Docker > Preferences > File Sharing.
...
 * `masters` matches the number of master nodes in the existing cluster,
 * `agents` matches the number of agent nodes in the existing cluster,
 * `public_agents` matches the number of public agent nodes in the existing cluster,
 * `destroy_on_error` must be `False`,
 * `destroy_on_success` must be `False`,
 * `files_to_copy_to_installer` must be `None` or `{}`,
+
+### DC/OS Installation
+
+`Cluster`s created by the ExistingCluster backend do not support installing DC/OS.
```
adamtheturtle: Backticks around ExistingCluster

src/dcos_e2e/node.py
```diff
@@ -140,15 +140,20 @@ def popen(
...
+        remote_path: Path,
+        user: str,
+    ) -> None:
         """
         Copy a file to this node.
 
         Args:
             local_path: The path on the host of the file to send.
             remote_path: The path on the node to place the file.
+            user: The name of the user used to send the file over SSH.
```
adamtheturtle: This makes it seem like it is a local user. Perhaps describe this as "the remote user". Optionally change `user` to `remote_user` but I don't like that because it conflicts with `run` a bit. Up to you 😄

adamtheturtle:
The CI fails with "fixture 'existing_cluster_backend' not found".

This is a PR that I've been wanting to see and we've both been planting the seeds for for weeks. Great to see it come together. Please re-submit after addressing the minor comments.

This is a conceptual change - I wonder if people will get stuck by running `run_integration_tests` before installing DC/OS, or similar. Let's keep an eye on that.

Before going further after merging this, I ask that you do a release and bump Enterprise DC/OS, to make sure that everything is smooth from a consumer's point of view.

src/dcos_e2e/backends/_existing_cluster.py
```diff
@@ -144,6 +127,46 @@ def __init__(  # pylint: disable=super-init-not-called
...
+    ) -> None:
+        """
+        Raises:
+            NotImplementedError: Raises `NotImplementedError` because it
+                is assumed that clusters created with the ExistingCluster
+                backend already have an installed instance of DC/OS
+                running on them.
+        """
+        message = (
+            'The ExistingCluster backend does not support installing DC/OS'
```
adamtheturtle: I know this is WIP but I spotted it - DC/OS is repeated "installing DC/OS DC/OS".

tests/test_harness.py
```diff
@@ -367,12 +352,12 @@ def test_false(
         preserved if there is no exception raised.
         """
         with Cluster(
-            build_artifact=oss_artifact,
             agents=0,
             public_agents=0,
             cluster_backend=cluster_backend,
             destroy_on_success=False,
         ) as cluster:
+            cluster.install_dcos_from_path(oss_artifact)
```
adamtheturtle: Is there any reason to install DC/OS in this test?

API.md
```diff
@@ -110,7 +98,47 @@ If set to `True`, the cluster is destroyed on exit if there is no exception rais
 
 ### Methods
 
-#### `run_integration_tests(pytest_command, env=None)`
+#### `install_dcos_from_url(build_artifact, extra_config=None, log_output_live=False)`
+
+Installs DC/OS on the given cluster using the DC/OS advanced installation method if supported
```
adamtheturtle: This repository follows  a practice inspired by http://rhodesmill.org/brandon/2012/one-sentence-per-line/ - one sentence per line for documentation source.

API.md
```diff
@@ -110,7 +98,47 @@ If set to `True`, the cluster is destroyed on exit if there is no exception rais
...
+The HTTP(S) URL string to a build artifact to install.
+
+##### `extra_config`
+
+Configuration variables to add to a base configuration.
+
+##### `log_output_live`
+
+If set to `True`, the output of the DC/OS installation process will be logged live.
+To see these logs in pytest tests, use the -s flag.
```
adamtheturtle: Wrap "-s" in backpacks.

API.md
```diff
@@ -131,6 +159,9 @@ cluster.run_integration_tests(
 )
 ```
 
+If set to `True`, the output of the `pytest_command` will be logged live.
+To see these logs in pytest tests, use the -s flag.
```
adamtheturtle: Backticks around -s.

src/dcos_e2e/backends/_base_classes.py
```diff
@@ -48,6 +38,28 @@ def __init__(
         """
 
     @abc.abstractmethod
+    def install_dcos_from_url(
+        self,
+        build_artifact: str,
+        extra_config: Dict[str, Any],
+        log_output_live: bool,
+    ) -> None:
```
adamtheturtle: Please document all `Args`.

src/dcos_e2e/backends/_base_classes.py
```diff
@@ -48,6 +38,28 @@ def __init__(
...
+
+    @abc.abstractmethod
+    def install_dcos_from_path(
+        self,
+        build_artifact: Path,
+        extra_config: Dict[str, Any],
+        log_output_live: bool,
+    ) -> None:
+        """
+        Install DC/OS from a build artifact passed as a file system `Path`.
```
adamtheturtle: Please document all `Args`.

src/dcos_e2e/backends/_docker/__init__.py
```diff
@@ -457,20 +422,56 @@ def __init__(  # pylint: disable=super-init-not-called,too-many-statements
...
         extra_config: Dict[str, Any],
+        log_output_live: bool,
     ) -> None:
+        """
+        Args:
+            build_artifact: The `Path` to a build artifact to install DC/OS
+                from.
+            extra_config: may contain extra installation configuration
+                variables that are applied on top of the default DC/OS
+                configuration of the docker backend.
```
adamtheturtle: `s/docker/Docker`

PR #314, opened 2017-11-14
Bump version

timaa2k:
Release the update from 2017.11.02.0 to 2017.11.14.0.


CHANGELOG.md
```diff
@@ -27,6 +28,8 @@
 
 ## Next
```
adamtheturtle: No need to change now, but I've traditionally used "Next" during development and not in releases.

PR #329, opened 2017-11-17
Split and improved wait_for_dcos utilizing dcos-test-utils

timaa2k:
In order to minimize the waiting time after starting up a DC/OS cluster with DC/OS E2E we need an appropriate procedure that helps to use the cluster as early as possible, thereby at least avoiding race conditions for ready states of user authentication and application deployment.


timaa2k:
DC/OS Security CLI tests passing:
https://github.com/mesosphere/dcos-security-cli/pull/94

adamtheturtle:
This looks great, thanks @timaa2k . However, the Security CLI test isn't quite enough - sorry I didn't think about this sooner - look at the note in `setup_cli` in that repository - that function is retried. To give us more confidence in this please adjust that function as necessary to remove the retry.

adamtheturtle:
I had a suspicion that `retry` could now be removed as a requirement in `requirements.txt`, and it can! I found a new linter we can use in CI to warn us of this.

```
> pip install pip_check_reqs
> pip-extra-reqs src/
Extra requirements:
requests in requirements.txt
retry in requirements.txt
```

Please move `requests` to `dev-requirements.txt` and remove `retry` from the requirements altogether.

dev-requirements.txt
```diff
@@ -1,3 +1,4 @@
+dcos-test-utils==0.1
```
adamtheturtle: `dev-requirements` are only ever installed *with* `requirements`. Therefore there shouldn't be anything here that is required for the library to be used.

adamtheturtle:
I updated this branch now that the new CI check is in `master`.

adamtheturtle:
As a follow-up, I think we now need a `README` entry example for DC/OS Enterprise.

timaa2k:
Updated security CLI test with comments fixed
https://github.com/mesosphere/dcos-security-cli/pull/94

adamtheturtle:
Great work @timaa2k . I'd like to leave it to you to release this and bump two consumers - the security CLI and DC/OS Enterprise.

PR #347, opened 2017-11-28
Fix vulture errors

adamtheturtle:


setup.py
```diff
@@ -28,8 +28,6 @@
 with open('README.md') as f:
     LONG_DESCRIPTION = f.read()
 
-TEST_UTILS_SHA = '45b76b1a961985343015996adeaa0bdd9128f4b3'
```
adamtheturtle: cc @timaa2k - Vulture caught this 😄

PR #354, opened 2017-12-04
Basic AWS Backend

timaa2k:
This PR adds a basic AWS backend to DC/OS E2E.
The semantic differences to the Docker backend are as follows.

- Nodes are currently not refreshed on each call, so in case of node failure E2E won't know.
- Copying files to the installer does not work.


tests/backends/test_aws.py
```diff
@@ -0,0 +1,72 @@
...
+            # See https://docs.pytest.org/en/latest/usage.html.
+            assert excinfo.value.returncode == 4
+
+    def test_install_dcos_from_path(
+        self,
+        aws_backend: ClusterBackend,
+        oss_artifact: Path,
+    ) -> None:
+        """
+        The Docker backend requires a build artifact in order
```
adamtheturtle: This docstring is not valid.

API.md
```diff
@@ -241,4 +247,8 @@ Copy a file to the node via SSH as the given user.
 
 #### `ip_address`
 
-The IP address of the node.
+The external IP address of the node.
+
+#### `private_ip_address`
+
+The DC/OS internal IP address of the node if different from the external IP address.
```
adamtheturtle: Could you add a little bit of information here as to what this is used for. Also, what do you think of having this required, rather than `None` ever - I guess it can be the same as `ip_address` where appropriate.

BACKENDS.md
```diff
@@ -77,6 +87,66 @@ On various platforms, the clock can get out of sync between the host machine and
...
+
+#### `instance_type`
+
+The [AWS instance type](https://aws.amazon.com/ec2/instance-types/) for all hosts of clusters created by this backend.
+
+#### `deploy_timeout`
+
+The lifespan measured in minutes for clusters created by this backend.
+
+#### `workspace_dir`
```
adamtheturtle: In the Docker backend, this is required because we make some really big files. Is this necessary for here? Maybe remove this and only add it if we need it.

setup.py
```diff
@@ -28,6 +28,8 @@
 with open('README.md') as f:
     LONG_DESCRIPTION = f.read()
 
+print(find_packages(where='src'))
```
adamtheturtle: Revert this

requirements.txt
```diff
@@ -1,8 +1,12 @@
...
 cryptography==2.1.4
 dcos-test-utils==0.1
+boto3==1.4.7
 docker==2.6.1
 paramiko==2.4.0
 PyYAML==3.12
 scp==0.10.2
+retrying==1.3.3
 urllib3==1.22
+setuptools==38.2.3
```
adamtheturtle: Why pin setuptools?

adamtheturtle: Ah because we use it in `setup.py`?

spelling_private_dict.txt
```diff
@@ -34,9 +36,9 @@ systemd
 tmp
 tmpfs
 travis
-typeshed
 url
 utils
 username
 vm
 xfail
+API
```
adamtheturtle: This file is alphabetically sorted apart from this one entry

src/dcos_e2e/backends/_aws/__init__.py
```diff
@@ -0,0 +1,424 @@
...
+
+from dcos_e2e.backends._aws import aws
+from dcos_e2e.backends._base_classes import ClusterBackend, ClusterManager
+from dcos_e2e.node import Node
+from dcos_test_utils import onprem, ssh_client
+from dcos_test_utils.helpers import Host
+
+log = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+# pylint: disable=logging-format-interpolation
```
adamtheturtle: Why?

src/dcos_e2e/backends/_aws/__init__.py
```diff
@@ -0,0 +1,424 @@
...
+
+        Raises:
+            NotImplementedError: `NotImplementedError` because the AWS
+                backend can be utilized more efficiently by installing DC/OS
+                from from a URL using the advanced installation method.
+        """
+        message = (
+            'The AWS backend does not support the installing DC/OS from '
+            'a build artifact on the file system. This is due a more '
+            'efficient method existing in `install_dcos_from_url`.'
```
adamtheturtle: Is the reason really that it is more efficient?

BACKENDS.md
timaa2k: It's useful to find the generated SSH key if you want to retrieve it for some purpose.

src/dcos_e2e/backends/_aws/__init__.py
```diff
@@ -0,0 +1,424 @@
...
+                from.
+            extra_config: This may contain extra installation configuration
+                variables that are applied on top of the default DC/OS
+                configuration of the AWS backend.
+            log_output_live: If `True`, log output of the installation live.
+        """
+        ssh_user = self._default_ssh_user
+
+        config = {
+            'cluster_name':
```
adamtheturtle: (Why) do we need a `cluster_name` ?

src/dcos_e2e/backends/_aws/__init__.py
timaa2k: Well, indirectly. If you create the build artifact locally then it is just as efficient.

src/dcos_e2e/backends/_aws/__init__.py
timaa2k: Not needed, true.

src/dcos_e2e/backends/_aws/__init__.py
```diff
@@ -0,0 +1,424 @@
...
+            self._cf_stack.stack_id,
+            'resolvers': ['8.8.8.8'],
+            'dns_search':
```
adamtheturtle: I'd really like a solid understanding of the differences between this and the equivalent in the Docker backend. If possible, I we should bring the creation of this configuration to the `Cluster` level and share it between backends.

src/dcos_e2e/node.py
```diff
@@ -30,6 +36,8 @@ def __init__(self, ip_address: IPv4Address, ssh_key_path: Path) -> None:
         """
         self.ip_address = ip_address
         self._ssh_key_path = ssh_key_path
+        self.private_ip_address = private_ip_address if private_ip_address \
```
adamtheturtle: Instead, I think it is better that `private_ip_address` is required. Happy to discuss.

adamtheturtle:
As discussed, let's take the following steps:

* PR to revert the change to Python 3.6, linking in the code to a JIRA issue to get Python 3.6 available
* PR to add a required `private_ip` to `Node`
* Experiment with calling DC/OS Launch code rather than copying it, where possible

adamtheturtle:
This should come with thorough documentation for the JSON templates.
It should maybe come with tests which at a high level assert the functionality desired from these templates.

Another option is to use the templates from DC/OS Launch.
As discussed, the problem with this is that we need direct access to Mesos and Marathon.

There are a few potential alternatives to the approach taken here.
These alternatives may reduce the maintenance burden of having copied templates between here and DC/OS Launch.
In no particular order:

* Change the templates in DC/OS Launch to meet the requirements of this project.
* Remove the reliance on direct access to Mesos and Metronome by expanding `dcos-diagnostics`.
* In Python, load the DC/OS Launch templates, make the required modifications, write the JSON.

.travis.yml
```diff
@@ -6,35 +6,41 @@ services:
...
+    - secure: "DCA/nGUV8IUrFg7SmL8K2WvLd3V/6JD7RINUyzBoKoADmJ9xvcQOI8Zm91//fJtiTNpEZOfOkFXYVQwGHAoPlgaS5/ENqbCV+4zn4RoS3q+WvuAaRR/mXj+z2Jzbgp+x9Pe1FTJroY521+Qqj1sjl9nQK24WOUPhlXe9bdFiRCNAK49r+dxa7dURVq0WyOcBk22psTxXpuBRfciV/dzun2llfdL0Ur9+enlMwj13HByMxR+TOPcsZ6ZXC4A8SxKKSpkT19lCTmfW34Ks0N6Gfqaad46eqfc97mwTyIaGW73iVLRsiI5sjBlXTlsdp2qPmn7WJih7BNOFdcoLNmqHW72akon+z5On2K5kwUqvobtrJ+d0n8hf+5HozZz4Tl6glByeMcWlcQgOqLZuH5scZK3U0rHOqeTsEDBlg45Y1CrTOCHWvetFgMwa9QQlmVNmnkWQ0B5fDuspW+sjqbeTMwwQk/2+eWYgi8vaR9N6gEouqUg5nZmpk3ppmY8NxHaVe1w5WWMOJyyws3T4f78eubGs3KLL2fcgLVhCWQNYa5zXXZRPG9d+kJsCCJFK7Bj8cIva31zbLTkTlA80xDxRYw33PvkWRKMP08zkyoo4MiYKvq42fpLuv7BbOksK1GosX+YvxyRK+yr0Uix62vNc4c6MwwLmtxWzJQEjOLEl4FU="
```
adamtheturtle: This is unwieldy, let's use secrets files.

https://docs.travis-ci.com/user/encrypting-files/#Caveat.

spelling_private_dict.txt
adamtheturtle: and I believe that it is case insensitive

bamarni:
Just stumbled upon this and I'd be interested in having it for testing on mac / windows nodes, is there any rough ETA for this PR?

timaa2k:
I abandoned this particular solution. I will work on a DC/OS launch backend for dcos-e2e as soon as I get to it (as soon as JP let‘s me do it.). There is a working backend in github.com/timaa2k/dcos-e2e@aws-backend that I currently use for resilience testing, however it‘s outdated I don‘t recommend to use it for anything serious.

I plan to work on this as soon as possible, but reproducing bugs in DC/OS is critical work at the moment.

BACKENDS.md
adamtheturtle: FWIW I've changed my perspective and in the `Docker` backend, having a workspace is now really useful - we can clean it up as part of destroying a cluster with the CLI.

adamtheturtle:
Closing as we are taking a different approach - calling DC/OS Launch functions as a library.

PR #356, opened 2017-12-05
disable fault domain script by default

mnaboka:
fault domain is now a required option in EE DC/OS.
This change disables the fault domains by default, however it is up to the user
to override it from a specific test. If we ever need to test a fault domain
in EE, the user might override this change from extra config.
https://jira.mesosphere.com/browse/DCOS-19702


adamtheturtle:
LGTM, thanks. I can re-run the tests if you are unable to, whenever you say that this is ready.

cc @timaa2k - you'll have to make this change in your AWS branch.

PR #359, opened 2017-12-07
Extend wait for dcos ee timeout

timaa2k:
Extending wait_for_dcos_ee timeout for waiting until the DC/OS CA cert can be fetched.


adamtheturtle:
There is a new required on EE fault domain option that we don't yet handle. I'll merge this as-is and fix that later. Thanks @timaa2k .

PR #363, opened 2017-12-08
Bring back dcos-diagnostics check prior to wait_for_dcos_oss/ee

timaa2k:
The PR should help to narrow down the wait_for_dcos/ee time to < 5 minutes by re-introducing the dcos-diagnostics check that was carried out in prior version of DC/OS E2E. The arbitrary waiting time after this check was set to 5 minutes back then.

### DIFF
--- a/src/dcos_e2e/cluster.py
+++ b/src/dcos_e2e/cluster.py
@@ -8,8 +8,11 @@ from contextlib import ContextDecorator
 from pathlib import Path
 from typing import Any, Dict, Iterable, List, Optional, Set
 
+import requests
 from dcos_test_utils.dcos_api import DcosApiSession, DcosUser
 from dcos_test_utils.helpers import CI_CREDENTIALS, session_tempfile
+from requests import codes
+from retry import retry
 from urllib3.exceptions import InsecureRequestWarning
 
 # Ignore a spurious error - this import is used in a type hint.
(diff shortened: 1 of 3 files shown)


src/dcos_e2e/cluster.py
```diff
@@ -90,6 +92,61 @@ def from_nodes(
...
+    @retry(
+        exceptions=(
+            subprocess.CalledProcessError,
+            ValueError,
+            requests.exceptions.ConnectionError,
+        ),
+        tries=500,
+        delay=5,
+    )
+    def _wait_for_adminrouter(
```
adamtheturtle: Why AdminRouter specifically?

src/dcos_e2e/cluster.py
```diff
@@ -90,6 +92,61 @@ def from_nodes(
...
+        Raises:
+            ValueError: If Admin Router is not capable yet of serving the CA
+                certificate for the DC/OS cluster.
+
+        """
+        diagnostics_args = [
+            '/opt/mesosphere/bin/dcos-diagnostics',
+            '--diag',
+            '||',
+            '/opt/mesosphere/bin/3dt',
```
adamtheturtle: Let's get rid of the 3dt check - no one is using such an old version of DC/OS with E2E.

src/dcos_e2e/cluster.py
timaa2k: Hmm, what a better name for: Wait until the CA certificate could be reached if this was an EE cluster or until Admin Router reports that the CA is not found if it is an OSS cluster.
Could call it wait_for_ca_cert_up_or_not_found maybe.

adamtheturtle: Instead, what do you think of splitting this up.

`wait_for_dcos_diagnostics` or maybe `wait_for_systemd_units` as one function.

Then `wait_for_ca_cert` in another. Then we could just not call `wait_for_ca_cert` in the OSS helper. And maybe we would keep your existing code to wait for a CA Cert.

timaa2k: Calling it _wait_for_dcos_startup now.

timaa2k: I totally agree. However I wanted to keep it as close as possible first to see if it even works.

adamtheturtle: Ok - well do you mind seeing if the tests pass and then rearranging and then resubmitting?

timaa2k: Yes, that's what I was going to do. I really want to see the EE tests pass first in Enterprise

src/dcos_e2e/cluster.py
```diff
@@ -127,36 +159,34 @@ def wait_for_dcos_ee(
...
             public_slaves=[str(n.ip_address) for n in self.public_agents],
             auth_user=DcosUser(credentials=credentials),
         )
 
-        warnings.simplefilter('ignore', InsecureRequestWarning)
-
-        ca_cert = api_session.get(
-            # We wait up to 30 minutes which is extraordinarily large
-            # but should avoid problems with 3 or 5 master counts.
+        response = enterprise_session.get(
```
adamtheturtle: I'd like a comment here and in the OSS equivalent on why `dcos-diagnostics` isn't enough. Ideally linking to JIRA issues.

src/dcos_e2e/cluster.py
```diff
@@ -127,36 +185,56 @@ def wait_for_dcos_ee(
...
+        # of services through HTTP.
+
+        # In the case of Enterprise DC/OS this method uses dcos-test-utils
+        # and superuser credentials to perform a superuser login that
+        # assure authenticating via CLI is working.
+
+        # Suggestion for replacing this with a DC/OS check for CLI login:
+
+        # In Enterprise
+        # DC/OS this could be replace by polling the login endpoint with
```
adamtheturtle: Why split line in the comment?

`s/replace/replaced`.

adamtheturtle:
Thanks @timaa2k - great work. Please make the minor change and merge - only wait for a linter to pass.

src/dcos_e2e/cluster.py
```diff
@@ -90,6 +90,36 @@ def from_nodes(
...
         )
 
+    @retry(
+        exceptions=(subprocess.CalledProcessError),
+        tries=500,
+        delay=5,
+    )
+    def _wait_for_dcos_diagnostics(
+        self,
+        log_output_live: bool = False,
```
adamtheturtle: There are no users of `log_output_live` and this is private. Please either remove it as an option or make it possible to pass through from `wait_for_dcos` methods and add tests for that.

adamtheturtle:
There are multiple technical words which need to be in the spelling wordlist.

src/dcos_e2e/cluster.py
```diff
@@ -98,6 +128,34 @@ def wait_for_dcos_oss(self) -> None:
...
+        # Current solution to guarantee the security CLI login:
+
+        # Try until one can login successfully with a long lived token
+        # (dirty hack in dcos-test-utils wait_for_dcos). This is to avoid
+        # having to simulate a browser that does the SSO flow.
+
+        # Suggestion for replacing this with a DC/OS check for CLI login:
+
+        # Determine and wait for all dependencies of the SSO OAuth login
+        # inside of DC/OS. This should include Admin Router, Zookeeper and
```
adamtheturtle: `s/Zookeeper/ZooKeeper.`

PR #375, opened 2018-01-09
Distinguish between host and DC/OS ip address at Node creation

timaa2k:
In order to support DC/OS cloud deployments with a dedicated subnet for DC/OS masters/agents we introduce the notion of a host and a DC/OS IP address when creating a new Node object. The host IP address is only required for connecting to the host via SSH. Therefore it can be hidden in the Node object. The DC/OS IP address that can be different from the host IP address is exposed as `ip_address` attribute. Universial DC/OS integration tests will only require the DC/OS IP accessible through the API. The host address is a implementation detail of the chosen backend.

### DIFF
--- a/API.md
+++ b/API.md
@@ -33,14 +33,15 @@
     - [`default_ssh_user`](#default_ssh_user)
 - [`dcos_e2e.node.Node`](#dcos_e2enodenode)
   - [Parameters](#parameters-1)
-    - [`ip_address`](#ip_address)
+    - [`host_ip_address`](#host_ip_address)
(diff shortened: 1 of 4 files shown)


API.md
```diff
@@ -198,14 +199,18 @@ The default SSH user to access cluster nodes.
 Commands can be run on nodes in clusters.
 
 ```python
-Node(ip_address, ssh_key_path)
+Node(host_ip_address, dcos_ip_address, ssh_key_path)
```
adamtheturtle: This should be a `CHANGELOG` entry.

PR #380, opened 2018-01-16
Expose public and private IP on Node object

timaa2k:
Recently the interface of the `Node` object was changed to take `host_ip_address` (public ip in DC/OS terms) and `dcos_ip_address` (private ip in DC/OS terms). `Node` then exposed only an `ip_address` attribute which is equivalent to `dcos_ip_address`.

The problem with this approach is that the DC/OS CLI setup requires the public ip to be readily available. Therefore we change the interface in such a way that the backend writer as well as the test writer must be aware of the types of ip addresses used within the DC/OS deployment method tied to a given backend. The `Node` object now exposes both, `public_ip_addess` and `private_ip_address`.

### DIFF
--- a/API.md
+++ b/API.md
@@ -33,15 +33,16 @@
     - [`default_ssh_user`](#default_ssh_user)
 - [`dcos_e2e.node.Node`](#dcos_e2enodenode)
   - [Parameters](#parameters-1)
-    - [`host_ip_address`](#host_ip_address)
-    - [`dcos_ip_address`](#dcos_ip_address)
+    - [`public_ip_address`](#public_ip_address)
+    - [`private_ip_address`](#private_ip_address)
     - [`ssh_key_path`](#ssh_key_path)
   - [Methods](#methods-1)
     - [`node.run(args, user, log_output_live=False, env=None)`](#noderunargs-user-log_output_livefalse-envnone)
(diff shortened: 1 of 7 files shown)


adamtheturtle:
I'm happy - @bamarni is this clear to you? Does it address your requirements? If so, please merge @timaa2k . Unfortunately Travis is having problems so we need to keep re-running builders. However, maybe we can add the tag to use the old/deprecated OS?

bamarni:
Thanks @timaa2k , that looks great. Just one concern about backwards compatibility, shall we still keep an undocumented property called `ip_address` so that it doesn't break for current users?

timaa2k:
The only consumers right now are the CLI and the dcos-enterprise integration tests.
We/I will update them on the next dcos-e2e bump. I think this is quite a substantial interface change that should be enforced while there are so little consumers.

adamtheturtle:
👍 @timaa2k .

@bamarni - we take the approach that we will put in the effort of updating consumers rather than the effort of maintaining backwards compatibility, at least for now.

@timaa2k - please re-run tests until they pass 😞

adamtheturtle:
Travis CI is hitting [backlog issues](https://www.traviscistatus.com/incidents/c1s4dlyxd4lf). I'm going to take the risk and merge - thanks @timaa2k for the work and thanks @bamarni for the feedback.

PR #384, opened 2018-01-24
Update the dcos-test-utils

alenkacz:
Improved messages for Metronome debugging


alenkacz:
@adamtheturtle

adamtheturtle:
Thank you @alena. You are now one of the top 5 contributors in terms of line changes to DC/OS E2E ;)

The tests are failing because this is from a fork.
Forks do not have access to secrets stored in Travis.
For example, in this repository, the secrets include the DC/OS Enterprise license.

Therefore, I have added you and all Mesosphere employees as collaborators to this repository. In future, you can and should use branches rather than forks for this repository.

However, in this case I will merge this as I see it as low-risk. We can always find that there is an issue later.

PR #385, opened 2018-01-24
[DCOS-20471] Escape shell commands

jongiddy:
Escape shell commands in the `Node.run` and `Node.popen` methods.  This means that passing 'hello world' in the list of args will be interpreted as a single argument on the remote node, rather than as two arguments separated by a space. This mimics the behavior of Python's `subprocess` module.

This is a better default because tests are typically written using simple parameters, and pass fine. Eventually a real value gets used that has a special character and something breaks, by now usually deep down in multiple layers of code.  On the other hand, when special characters are meant to be interpreted as special characters, tests break quickly, and a solution is identified.

To provide that solution, a `shell` parameter has been added, similar to the `subprocess` parameter. This provides behavior more like the old behavior, with spaces and other special characters being interpreted using shell semantics. The main difference is that the `Node` methods still expect a list of strings, and do not accept a plain string.

The `dcos-enterprise` `test-e2e` tests all use literal values, so will not break with these changes. The only exception is the test that I have in a Pull Request, which will need to be changed when this is added.  However, I am not sure whether this e2e code is used anywhere else?

### DIFF
--- a/tests/test_node.py
+++ b/tests/test_node.py
@@ -47,19 +47,16 @@ class TestNode:
     Tests for interacting with cluster nodes.
     """
 
-    def test_run(
+    def test_run_literal(
         self,
-        caplog: LogCaptureFixture,
         dcos_cluster: Cluster,
     ) -> None:
         """
-        It is possible to run commands as the given user and see their output.
+        When shell=False, preserve arguments as literal values.
         """
         (master, ) = dcos_cluster.masters
         default = dcos_cluster.default_ssh_user
 
-        # Arguments are quoted - spaces and special characters are preserved
-        # as literal values
         echo_result = master.run(
             args=['echo', 'Hello, ', '&&', 'echo', 'World!'], user=default
         )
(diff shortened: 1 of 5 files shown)


adamtheturtle:
> However, I am not sure whether this e2e code is used anywhere else?

At least the [security CLI](https://github.com/mesosphere/dcos-security-cli) and the dcos licensing project.

tests/test_node.py
```diff
@@ -58,11 +58,44 @@ def test_run(
         (master, ) = dcos_cluster.masters
         default = dcos_cluster.default_ssh_user
 
-        echo_result = master.run(args=['echo', '$USER'], user=default)
+        # Arguments are quoted - spaces and special characters are preserved
```
adamtheturtle: Note that all of the tests in this module use a shared cluster. If the reason that you put all of these assertions into one test was to save time, you can split them up. If not, I'm OK 😄

adamtheturtle: That is, I would probably go for `def test_shell_false` and `def test_shell_true` or similar. Maybe even `class TestShell`.

adamtheturtle:
Thanks @jongiddy - I have a preference that you address https://github.com/mesosphere/dcos-e2e/pull/385#discussion_r163589555 and resubmit but feel free to choose not to, assuming that tests pass.

timaa2k:
Cool effort Jon! That's a very valuable E2E feature for me right now. 💯

PR #387, opened 2018-01-24
Node string representation

timaa2k:
This PR adds a custom `string` representation for the `Node` object, listing the corresponding `public_ip_address` and `private_ip_address`.


adamtheturtle:
Thanks @timaa2k . Please add a test and resubmit.

PR #401, opened 2018-02-01
[WIP] Add a DC/OS Launch backend

bamarni:
This adds a DC/OS launch backend, it is currently not very generic as it was made to satisfy my specific needs in DC/OS licensing integration tests.

Hopefully we can start to work from this.

TODO : 
- [ ] Add `dcos_launch` requirement
- [ ] Make it more generic

### DIFF
--- a/src/dcos_e2e/backends/__init__.py
+++ b/src/dcos_e2e/backends/__init__.py
@@ -3,12 +3,14 @@ Classes to allow backend-specific configuration for cluster backend types.
 """
 
 from ._base_classes import ClusterBackend, ClusterManager
+from ._dcos_launch import DCOSLaunch
 from ._docker import Docker
 from ._existing_cluster import ExistingCluster as _ExistingCluster
 
 __all__ = [
     'ClusterBackend',
     'ClusterManager',
+    'DCOSLaunch',
     'Docker',
     '_ExistingCluster',
 ]
--- a/src/dcos_e2e/backends/_dcos_launch/__init__.py
+++ b/src/dcos_e2e/backends/_dcos_launch/__init__.py
@@ -1,0 +1,174 @@
+"""
+A DC/OS Launch backend for DC/OS E2E.
+"""
+import uuid
+import sys
+
+from ipaddress import IPv4Address
+from pathlib import Path
+from typing import Any, Dict, Set
+
+from dcos_e2e.backends import ClusterBackend, ClusterManager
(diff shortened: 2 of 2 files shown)


adamtheturtle:
We should discuss with @mellenburg how we should take this further. My initial thought is that we should not have DC/OS Launch backend. Instead, we should have e.g. an AWS backend, and that just so happens to be powered by DC/OS Launch. Why does a user care that it is powered by DC/OS Launch?

cc @timaa2k

bamarni:
I assume there are (or will be) different kind of providers available through dc/os launch (like gcp), and maybe also different aws setups?

As such I think it makes sense to have a raw dc/os launch backend, accepting a dc/os launch config as a dict, or a 1:1 parameter argument mapping with dc/os launch config.

bamarni:
We could then have an AWS backend which extends DCOSLaunch.

timaa2k:
I don't see the difference between GCE and AWS for the use cases that we have, they are not specific to a cloud platform (launching onprem in the cloud anyway). The reason we opted out for DC/OS launch as a backend is that E2E is not bothered with breakage of launching of cloud clusters in the future. If we can preserve the clean interface of E2E while providing only one option to launch a cloud cluster and avoid having to keep it updated on DC/OS launching procedure changes that is a very good start.

adamtheturtle:
> As such I think it makes sense to have a raw dc/os launch backend, accepting a dc/os launch config as a dict, or a 1:1 parameter argument mapping with dc/os launch config.

Perhaps, but I'd be interested in seeing a use case first.

In my mind, we have a DC/OS Launch backend perhaps, but it is not public. Similar to how we have `ExistingCluster` but it is only accessible via `from_nodes`.

timaa2k:
The only use case I can think of is for launching a cloud formation template.
But we could have an AWS onprem and an AWS CloudFormation backend for that both use the internal DC/OS Launch backend initially.

bamarni:
I'm fine with hiding dc/os launch if you'd rather see it as an implementation detail.

Otherwise one limitation that I had seen is that dc/os launch doesn't run on windows, so I'm spinning-up a cluster on linux and then use an existing cluster backend on windows e2e tests. I did a helper function to generate a dc/os launch config. Do you think this should still be like this?

adamtheturtle:
I'm going to close this and we will keep it in mind for when we can focus on a cloud backend. Thanks @bamarni !

PR #421, opened 2018-02-07
Consolidate child process handling code

jgehrcke:
This is no functional improvement, the code works as-is, but for a different reason that originally intended / commented. This attempts to make the code more expressive and correct.

Commit message:

```
`communciate()` calls `wait()` behind the scenes, but only
`wait()` is actually what is required here to populate
the `process.returncode` object (the `wait()` system call
is doing exactly that: wait for process termination and
reap the zombie state and read the exit status). With that
change `poll()` is not actually required because the exit
status is known to be known.
```

I am throwing this over the fence a bit, let's discuss on Slack if desired :-).

### DIFF
--- a/src/dcos_e2e/_common.py
+++ b/src/dcos_e2e/_common.py
@@ -64,7 +64,7 @@ def run_subprocess(
                     stdout += line
                 # stderr/stdout are not readable anymore which usually means
                 # that the child process has exited. However, the child
-                # process has not been wait()et for yet, i.e. it has not yet
+                # process has not been wait()ed for yet, i.e. it has not yet
                 # been reaped. That is, its exit status is unknown. Read its
                 # exit status.
                 process.wait()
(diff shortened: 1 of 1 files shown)


src/dcos_e2e/_common.py
```diff
@@ -62,10 +62,12 @@ def run_subprocess(
                         line.rstrip().decode('ascii', 'backslashreplace')
                     )
                     stdout += line
-                # Without this, `.poll()` will return None on some
-                # systems.
-                # See https://stackoverflow.com/a/33563376.
-                process.communicate()
+                # stderr/stdout are not readable anymore which usually means
+                # that the child process has exited. However, the child
+                # process has not been wait()et for yet, i.e. it has not yet
```
adamtheturtle: Note for me - `et` > `ed`

jongiddy:
This looks good to me. The StackOverflow issue seems to refer to a full buffer that occurs when the pipe is not drained. This is not a problem in this case.

Having just reviewed the `subprocess` documentation, I notice that return code can be negative when a signal terminates the process. Not directly related to this change, but should `if process.returncode > 0:` use `!= 0` instead?

jgehrcke:
> Not directly related to this change, but should if process.returncode > 0: use != 0 instead?

Yes! :)

adamtheturtle:
Thanks for the change and the lesson, both!

PR #459, opened 2018-02-13
Default ssh user in node

timaa2k:
This PR moves the `default_ssh_user` from being a public attribute of a cluster to a public attribute of each `Node` of the cluster. It also sets the `default_ssh_user` to be the default by which SSH commands are issues using `Node.run`, `Node.popen` and `Node.sendfile`. Therefore it is no longer required to pass a `user` to those functions.


src/dcos_e2e/node.py
```diff
@@ -150,6 +154,9 @@ def run(
             subprocess.CalledProcessError: The process exited with a non-zero
                 code.
         """
+        if not user:
```
adamtheturtle: This uses the default user if `user` is set to `''`. Is that a problem?

CHANGELOG.rst
```diff
@@ -6,6 +6,8 @@ Changelog
 Next
 ----
 
+-  Backwards incompatible change: Move ``default_ssh_user`` parameter from ``Cluster`` to ``Node``.
+   The ``default_ssh_user`` is now for ``Node.run``, ``Node.popen`` and ``Node.sendfile`` if the ``user`` is not supplied.
```
adamtheturtle: ``send_file`` with an underscore.

"is now for" I guessed "is now used for".

src/dcos_e2e/node.py
```diff
@@ -130,7 +133,8 @@ def run(
 
         Args:
             args: The command to run on the node.
-            user: The username to SSH as.
+            user: The username to SSH as. If unset then the `default_ssh_user`
```
adamtheturtle: ``` s/unset/``None`` ```

tests/test_dcos_e2e/backends/test_docker.py
```diff
@@ -161,10 +159,7 @@ def test_default(self) -> None:
             public_agents=0,
         ) as cluster:
             (master, ) = cluster.masters
-            node_distribution = self._get_node_distribution(
-                node=master,
-                default_ssh_user=cluster.default_ssh_user,
-            )
+            node_distribution = self._get_node_distribution(node=master, )
```
adamtheturtle: Trailing comma in the brackets

adamtheturtle: Here and below

adamtheturtle:
```python
        # Logically a SSH user should be part of a `Node`.
        # However with the DC/OS config there is a notion of a SSH user for
        # an entire DC/OS cluster, which in our case maps to a single SSH user
        # for every cluster created with the corresponding backend. Maybe
        # we're better off making this a `default_ssh_user` of a
        # `ClusterManager` instead.
        self._default_ssh_user = cluster_backend.default_ssh_user
```

That's part of `_docker/__init__.py`.
I'll leave the choice to you but this should be removed.

adamtheturtle:
The coverage is failing because there is no coverage for the invisible `else` case of `if user is None:`.

src/dcos_e2e/node.py
```diff
@@ -23,22 +23,25 @@ def __init__(
...
     ) -> None:
         """
         Args:
             public_ip_address: The public IP address of the node.
             private_ip_address: The IP address used by the DC/OS component
                 running on this node.
+            default_ssh_user: The default username to use for SSH connections.
             ssh_key_path: The path to an SSH key which can be used to SSH to
-                the node as the ``Cluster.default_ssh_user``.
+                the node as the `default_ssh_user` user.
```
adamtheturtle: Double backticks around `default_ssh_user` for monospace - this is generated using Rst and single backticks are italics there.

src/dcos_e2e/node.py
```diff
@@ -226,24 +237,28 @@ def send_file(
             local_path: The path on the host of the file to send.
             remote_path: The path on the node to place the file.
             user: The name of the remote user to send the file via
-                secure copy.
+                secure copy. If `None` the ``default_ssh_user`` is
```
adamtheturtle: Two backticks around `None` (you can change after tests pass and merge anyways).

PR #558, opened 2018-02-28
[DCOS_OSS-2177] Add instructions for deploying on Windows, with supporting Vagrantfile

jongiddy:
Add instructions for running DC/OS E2E on Windows.  The only way we could get it to work was using Vagrant.  This works similarly to `dcos-docker`'s Vagrant setup, but I have rewritten it to use Ubuntu 16.04. Because Ubuntu is generally more up-to-date than CentOS, it is simpler to install Python 3.5 and newer versions of Docker.

Having the Docker build in this repo means the users have to wait for the VM software to install, but it also provides the flexibility to change the setup. I've also tried to explain why everything is in there, to avoid any "magic that just makes it work".

The description for the Windows installation is applicable to Mac and Linux as well. Just needs different commands to download the Vagrantfile and to setup the routing.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -35,4 +35,61 @@ The latest release is |release|.
 
     pip install git+https://github.com/mesosphere/dcos-e2e.git@master
 
+
+Windows
+~~~~~~~
+
+The only supported way to run DC/OS E2E on Windows is using Vagrant and VirtualBox.
+
+- Ensure Virtualization and VT-X support is enabled in your PC's BIOS.
+- Disable Hyper-V virtualisation.
+- Install VirtualBox and VirtualBox Extension Pack.
+- Install Vagrant.
+- Install the Vagrant plugin for persistent disks:
+
+.. code:: ps1
+
+    vagrant plugin install vagrant-persistent-storage
+
+- Optionally install the Vagrant plugins to cache package downloads and keep guest additions updates:
+
+.. code:: ps1
+
+        vagrant plugin install vagrant-cachier
+        vagrant plugin install vagrant-vbguest
+
+- Start Powershell and download the E2E ``Vagrantfile`` to a directory containing a DC/OS installer file:
+
+.. code:: ps1
+
+    ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/mesosphere/dcos-e2e/master/vagrant/Vagrantfile')) | Set-Content -LiteralPath Vagrantfile
+
+- Start the virtual machine and login:
+
+.. code:: ps1
+
+    vagrant up
+    vagrant ssh
+
+You can now run ``dcos-docker`` commands.
+
+To connect to the cluster nodes from the Windows host (e.g. to use the DC/OS web interface), in PowerShell Run as Administrator, and add the Virtual Machine as a gateway:
+
+.. code:: ps1
+
+    route add 172.17.0.0 MASK 255.255.0.0 192.168.18.2
+
+To shutdown, logout of the virtual machine shell, and destroy the virtual machine and disk
+
+.. code:: ps1
+
+    vagrant destroy
+
+The route will be removed on reboot. You can manually remove the route in PowerShell Run as Administrator using:
+
+.. code:: ps1
+
+    route delete 172.17.0.0
+
+
 .. _Homebrew: https://brew.sh
--- a/vagrant/Vagrantfile
+++ b/vagrant/Vagrantfile
@@ -1,0 +1,81 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+
+# Provisioning script to install Docker and dcos-e2e
+$provision_script = <<SCRIPT
+#!/bin/bash
+
+set -o errexit -o nounset -o pipefail -x
+
+# Docker CE installation instructions
+apt-get update
+apt-get install -y apt-transport-https ca-certificates curl software-properties-common
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
+apt-key fingerprint 0EBFCD88 | grep -c Docker
+add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+apt-get update
+apt-get install -y docker-ce=17.12.0~ce-0~ubuntu
+
+# Allow user to run Docker commands
+usermod -a -G docker ubuntu
+
+# Install dcos-e2e
+apt-get install -y python3-pip
+pip3 install git+https://github.com/mesosphere/dcos-e2e.git@master
+
+# Accept IP forwarding to Docker containers.  On the VM host, set this VM guest's IP (192.168.18.2)
+# as the gateway for routing 172.17.0.0/16.  This allows the VM host to connect to Docker containers,
+# to use the DC/OS web interface, for example.
+iptables -I DOCKER-USER ! -i docker0 -o docker0 -j ACCEPT
+SCRIPT
+
+# Provisioning script to start user in /vagrant directory
+$user_script = <<SCRIPT
+#!/bin/bash
+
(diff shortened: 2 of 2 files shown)


docs/source/installation.rst
```diff
@@ -35,4 +35,61 @@ The latest release is |release|.
 
     pip install git+https://github.com/mesosphere/dcos-e2e.git@master
 
+
+Windows
```
adamtheturtle: Please can you link to this, or mention it, in `README.rst`.

adamtheturtle:
This deserves a `CHANGELOG.rst` entry!

adamtheturtle:
What happens when you run `dcos-docker web` inside the VM?

docs/source/installation.rst
```diff
@@ -35,4 +35,61 @@ The latest release is |release|.
...
+- Start the virtual machine and login:
+
+.. code:: ps1
+
+    vagrant up
+    vagrant ssh
+
+You can now run ``dcos-docker`` commands.
+
+To connect to the cluster nodes from the Windows host (e.g. to use the DC/OS web interface), in PowerShell Run as Administrator, and add the Virtual Machine as a gateway:
```
adamtheturtle: As `docs-docker web` is not available, perhaps say how exactly to find the web UI address - with `dcos-docker inspect` on the VM.

docs/source/installation.rst
```diff
@@ -35,4 +35,61 @@ The latest release is |release|.
...
+
+.. code:: ps1
+
+    route add 172.17.0.0 MASK 255.255.0.0 192.168.18.2
+
+To shutdown, logout of the virtual machine shell, and destroy the virtual machine and disk
```
adamtheturtle: I think end this with `:`.

docs/source/installation.rst
```diff
@@ -35,4 +35,61 @@ The latest release is |release|.
...
+    ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/mesosphere/dcos-e2e/master/vagrant/Vagrantfile')) | Set-Content -LiteralPath Vagrantfile
+
+- Start the virtual machine and login:
+
+.. code:: ps1
+
+    vagrant up
+    vagrant ssh
+
+You can now run ``dcos-docker`` commands.
```
adamtheturtle: ... and run code which uses the library. I'd change ``` ``dcos-docker`` ``` to `` :doc:`cli` `` - or at least try that to see if it is better, and maybe then add ``` :doc:`library` ```.

docs/source/installation.rst
```diff
@@ -35,4 +35,61 @@ The latest release is |release|.
...
+
+To connect to the cluster nodes from the Windows host (e.g. to use the DC/OS web interface), in PowerShell Run as Administrator, and add the Virtual Machine as a gateway:
+
+.. code:: ps1
+
+    route add 172.17.0.0 MASK 255.255.0.0 192.168.18.2
```
adamtheturtle: Can `-p` be used for persistence across reboots? Maybe document this if so.

docs/source/installation.rst
```diff
@@ -35,4 +35,61 @@ The latest release is |release|.
 
     pip install git+https://github.com/mesosphere/dcos-e2e.git@master
 
+
+Windows
+~~~~~~~
+
```
adamtheturtle: Please document that this gets you `master` in the VM.

There are trade offs and various options. From what I can see those options are:

* No dcos-e2e installation. You install your preferred version yourself.
* master
* The latest released version

Let's at least document the choice here, and note to ourselves that we might want to change this. I find it a little strange that the Homebrew recipe gets us the latest released version, and this gets us "master".

vagrant/Vagrantfile
```diff
@@ -0,0 +1,81 @@
...
+Vagrant.configure("2") do |config|
+
+  if Vagrant.has_plugin?("vagrant-cachier")
+    config.cache.scope = :box
+  end
+  if Vagrant.has_plugin?('vagrant-vbguest')
+    config.vbguest.auto_update = true
+  end
+
+  config.vm.box = "ubuntu/xenial64"
```
adamtheturtle: What do you think of using a box with Docker already installed? https://github.com/William-Yeh/docker-enabled-vagrant ?

docs/source/installation.rst
jongiddy: 👍

vagrant/Vagrantfile
jongiddy: The thing I like about installing the software in the Vagrantfile is that when you do find yourself in a strange and unexpected variation of OS/network setup, it's quick to modify the configuration to get around problems.

I'd suggest a good follow-up could be to add PowerShell and bash scripts to build the box once, and save it locally. This gives the speed of a pre-built box with the flexibility of having the `Vagrantfile` right there.

docs/source/installation.rst
jongiddy: @adamtheturtle Do you manually update that for each release? Are there release instructions to manage this?

adamtheturtle: @jongiddy the release script in the `admin` directory updates the Homebrew recipe.

docs/source/installation.rst
jongiddy: Maybe `dcos-docker web` could display a message "Opening browser to http://172.17.0.2/". Then, if it fails (and looking at the `click` code, that's not unlikely), the user can deal with it?

docs/source/installation.rst
```diff
@@ -35,4 +35,61 @@ The latest release is |release|.
...
+The only supported way to run DC/OS E2E on Windows is using Vagrant and VirtualBox.
+
+- Ensure Virtualization and VT-X support is enabled in your PC's BIOS.
```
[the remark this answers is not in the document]
jongiddy: 👍

docs/source/installation.rst
adamtheturtle: Maybe - let's find what it does at least first. This can be a follow-up if you prefer.

docs/source/installation.rst
adamtheturtle: Feel free to ignore this / do it as a follow-up as you see fit.

adamtheturtle:
We can merge this without tests but we need the linter to pass.

First issue:

```
lists of files in version control and sdist do not match!
missing from sdist:
  vagrant
  vagrant/Vagrantfile
make: *** [check-manifest] Error 1
The command "make lint" failed and exited with 2 during .
```

Use `setup.cfg` to say that we don't want to distribute these files to people `pip` installing the package.

docs/source/installation.rst
jongiddy: It is done - it will now use the latest release (at the time of downloading the Vagrantfile) and allows overriding using an environment variable.

docs/source/installation.rst
```diff
@@ -35,4 +35,68 @@ The latest release is |release|.
...
+
+.. code:: ps1
+
+    $env:DCOS_E2E_REF = "master"
+
+- Start the virtual machine and login:
+
+.. code:: ps1
+
+    vagrant up
```
adamtheturtle: This requires the user to have the Vagrantfile.
The instructions in this document do not assume that the user is in a clone of the repository.

Instead, instruct that the user downloads the Vagrantfile.
I'm not sure which tools are available on Windows by default, but for the release process we do:

```
curl https://raw.githubusercontent.com/mesosphere/dcos-e2e/master/admin/release.sh
```

jongiddy: The download is two bullet points up.

adamtheturtle: My mistake - sorry!

vagrant/Vagrantfile
adamtheturtle: 😄 I'll leave it to you to decide what's best. Feel free to make any JIRA issues you like for this.

PR #612, opened 2018-03-20
Dump command output to file

timaa2k:
This PR adds a `dump_output` method to the `Node` object. The method runs a command using `Node.run` and dumps the output to a file with a name generated like so: `{private_ip}_{public_ip}_{args}_{datetime}.log`

There are two tests for this functionality, one especially for a custom operating system user.


adamtheturtle:
@timaa2k and I discussed on Slack that this should be a helper in the resilience tests, at least at first. Then we can consider later adding it to DC/OS E2E.

This is to keep the API footprint of E2E small.

adamtheturtle:
Closing as per the comments on https://jira.mesosphere.com/browse/DCOS-21726.

PR #618, opened 2018-03-21
base-docker:  fix 17.12.1-ce dockerfile

gpaul:
The curl was failing with 404 Not Found, causing the entire RUN instruction to fail.


adamtheturtle:
Thanks @gpaul - if this works, great.

What I'd like to know:

* How come this does not get caught in our tests?
* Should the `|| true` removals be applied to other, similar `Dockerfile`s in this directory?
* (How) could we change the code so that we can see this error more easily?

adamtheturtle:
I'd also rather just remove the bash completion thing.
This has caused us problems in the past, and costs us reliability, bandwidth and space (minimal but non-0).

I'm going to work further on this and thank @gpaul for doing the hard work!

adamtheturtle:
Closing in favour of a different approach as tracked by https://jira.mesosphere.com/browse/DCOS_OSS-2288

PR #722, opened 2018-04-11
Aws backend dcos 1.9 workaround

timaa2k:
Workaround for the 1.9 bug described here: [DCOS_OSS-847](https://jira.mesosphere.com/browse/DCOS_OSS-847)

Due to a bug in < 1.9.7 `ip_detect_public_filename` has no effect when supplied in the `config.yaml`.
Another unintended behavior, overwriting `ip_detect_public_contents` with a functioning `ip-detect-public` script, can work around this issue until the bug fix is merged into 1.9.8. The workaround also enables us to launch DC/OS < 1.9.7 after the fix was merged. If then the other unintended behavior is fixed someday, and `ip_detect_public_contents` it may not work anymore, because dcos-launch does not add `ip_detect_public_filename` automatically with its pre-defined script if `ip_detect_public_contents` is present. Then we need to take action to differentiate DC/OS versions or remove the workaround and stop supporting < 1.9.7.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,8 @@ Changelog
 Next
 ----
 
+- The AWS backend now supports DC/OS 1.9.
+
 2018.04.11.0
 ------------
 
--- a/src/dcos_e2e/backends/_aws/__init__.py
+++ b/src/dcos_e2e/backends/_aws/__init__.py
@@ -182,9 +182,8 @@ class AWSCluster(ClusterManager):
         # dcos-launch config to pass the config validation step.
         launch_config['dcos_config'] = {
             'cluster_name': unique,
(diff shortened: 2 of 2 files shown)


adamtheturtle:
Please:

* Address my tiny comment
* Add a Changelog entry like "The ``AWS`` backend now supports DC/OS 1.9".
* Confirm that you have tested this on a 1.9 cluster
* Wait until AWS tests on CI are done

adamtheturtle:
and thanks!

timaa2k:
Tested multiple times with the CI TestRunIntegration tests.
Proof of [1.9.x resilience tests in Teamcity](https://teamcity.mesosphere.io/viewLog.html?buildId=1032501&buildTypeId=DcOs_Enterprise_Test_DockerBased_E2eFaultInjection19&tab=buildLog) (past cluster start already.)

PR #738, opened 2018-04-18
Set the log level specifically for the log_output_live logger

timaa2k:
This change sets the log level to DEBUG specifically for the `LOGGER` that is used to log the live output.
This avoids affecting other loggers by not requiring to modify the `logging.basicConfig`.

### DIFF
--- a/src/dcos_e2e/_common.py
+++ b/src/dcos_e2e/_common.py
@@ -21,7 +21,7 @@ def get_logger(name: str) -> logging.Logger:
     """
     # This gets the root logger with a ``logging.lastResort``
     # StreamHandler that logs on WARNING level.
-    logger = logging.getLogger(__name__)
+    logger = logging.getLogger(name)
 
     # Do this if a logger is called for the first time
     if not logger.hasHandlers():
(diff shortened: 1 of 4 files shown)


src/dcos_e2e/_common.py
```diff
@@ -7,8 +7,8 @@
 from subprocess import PIPE, STDOUT, CompletedProcess, Popen
 from typing import Dict, List, Optional, Union
 
-logging.basicConfig(level=logging.DEBUG)
 LOGGER = logging.getLogger(__name__)
+LOGGER.setLevel(logging.DEBUG)
```
adamtheturtle: Is there any use of `dcos_e2e` for which having this line makes a difference?

src/dcos_e2e/_common.py
```diff
@@ -7,8 +7,44 @@
...
+
+    Returns:
+        See :py:class:`logging.Logger`.
+    """
+    # This gets the root logger with a ``logging.lastResort``
+    # StreamHandler that logs on WARNING level.
+    logger = logging.getLogger(name)
+
+    # Do this if a logger is called for the first time
+    if not logger.hasHandlers():
```
adamtheturtle: Will this get 100% coverage? That is, does the test suite cover both `logger.hasHandlers() is True` and `logger.hasHandlers() if False`?

adamtheturtle:
Please add a changelog entry.

adamtheturtle:
Please address the minor comments and then I'll be happy to merge once the relevant tests have passed.

PR #826, opened 2018-05-08
Fix CLI --sync-dir example

jongiddy:
Updated this example to use the actual option name and match another example.


adamtheturtle:
🥂

PR #892, opened 2018-05-29
Setuptools greater equals

adamtheturtle:


adamtheturtle:
@aweigold No worries! We're going to think soon about being a little more flexible with the requirements as a whole. Feel free to get in touch / make an issue if you have any preferences.

PR #1030, opened 2018-06-26
Add custom network option for the Docker backend

timaa2k:
In order take ownership of the network that Docker containers are connected to with the Docker backend we add the functionality to supply a custom Docker network when creating a new Docker backend. If the `network` parameter is set, all `Cluster`s created with the given backend will be connected to this network. If no `network` parameter is set, all `Cluster`s will be connected to the default `docker0` bridge network as before.


src/dcos_e2e/_node_transports/_docker_exec_transport.py
```diff
@@ -256,3 +244,21 @@ def send_file(
             ssh_key_path=ssh_key_path,
             public_ip_address=public_ip_address,
         )
+
+
+def _get_container_from_ip_address(ip_addr: Union[IPv4Address, IPv6Address],
+                                   ) -> docker.models.containers.Container:
+    """
+    Return the container which represents the given ``node``.
```
adamtheturtle: This is not a correct docstring.

tests/test_dcos_e2e/backends/docker/test_docker.py
```diff
@@ -377,7 +382,80 @@ class TestNetworks:
...
+            assert custom_network_ip == str(master.public_ip_address)
+            assert custom_network_ip == str(master.private_ip_address)
+
+    @pytest.mark.parametrize('transport', list(Transport))
+    def test_transport(
+        self,
+        docker_network: Network,
+        transport: Transport,
+        tmpdir: local,
+    ) -> None:
```
adamtheturtle: Test needs a docstring.

src/dcos_e2e/backends/_docker/_containers.py
```diff
@@ -108,6 +109,8 @@ def start_dcos_container(
         docker_version: The Docker version to use on the node.
         docker_storage_driver: The storage driver to use for Docker on the
             node.
+        network: The network a container is connected to other than the default
```
adamtheturtle: For consistency "The Docker network to connect the container to...".

src/dcos_e2e/backends/_docker/__init__.py
```diff
@@ -529,13 +536,15 @@ def destroy_node(self, node: Node) -> None:
...
-        ]
-        container.stop()
-        container.remove(v=True)
+        for container in containers:
+            networks = container.attrs['NetworkSettings']['Networks']
+            for net in networks:
+                if networks[net]['IPAddress'] == str(node.public_ip_address):
+                    node_container = container
+                    break
+        assert node_container
```
adamtheturtle: No need for this assertion - the next line will fail if it is `None`.

tests/test_dcos_e2e/backends/docker/test_docker.py
```diff
@@ -377,7 +382,80 @@ class TestNetworks:
...
+        Return a Docker network.
+        """
+        client = docker.from_env(version='auto')
+        ipam_pool = docker.types.IPAMPool(
+            subnet='172.28.0.0/16',
+            iprange='172.28.0.0/24',
+            gateway='172.28.0.254',
+        )
+        network = client.networks.create(
+            name='dcos-e2e-network-{random}'.format(random=uuid.uuid4()),
```
adamtheturtle: Comment that this name starts with `dcos-e2e` because that matches the network name prefix in the `Makefile`.

adamtheturtle:
Great work @timaa2k !

PR #1036, opened 2018-06-28
Support custom AWS instance types

bamarni:
The default `m4.large` is not enough for CLI integration tests.

https://jira.mesosphere.com/browse/DCOS_OSS-3696

### DIFF
--- a/src/dcos_e2e/backends/_aws/__init__.py
+++ b/src/dcos_e2e/backends/_aws/__init__.py
@@ -27,6 +27,7 @@ class AWS(ClusterBackend):
 
     def __init__(
         self,
+        aws_instance_type: str = 'm4.large',
         aws_region: str = 'us-west-2',
         admin_location: str = '0.0.0.0/0',
         linux_distribution: Distribution = Distribution.CENTOS_7,
@@ -39,6 +40,8 @@ class AWS(ClusterBackend):
         Args:
             admin_location: The IP address range from which the AWS nodes can
                 be accessed.
+            aws_instance_type: The AWS instance type to use.
+                See `Instance types`_.
             aws_region: The AWS location to create nodes in. See
                 `Regions and Availability Zones`_.
             linux_distribution: The Linux distribution to boot DC/OS on.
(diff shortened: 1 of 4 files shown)


adamtheturtle:
Thank you @bamarni !

As mentioned by @timaa2k on Slack, http://dcos-e2e.readthedocs.io/en/latest/aws-backend.html#ec2-instance-types should be updated.

Please also add a changelog entry under `Next` at https://github.com/dcos/dcos-e2e/blob/master/CHANGELOG.rst#next.

Beyond that, I'm happy to merge.

The CI uses Travis CI secrets and so cannot be run from a fork. Are you able to recreate this PR from a branch of `dcos/dcos-e2e` rather than your own branch?

tests/test_dcos_e2e/backends/aws/test_aws.py
```diff
@@ -32,6 +32,12 @@ def test_admin_location(self) -> None:
         """
         assert AWS().admin_location == '0.0.0.0/0'
 
+    def test_aws_instance_type(self) -> None:
```
adamtheturtle: This test will not pass.

In particular, there is no `aws_instance_type` attribute set in the `AWS.__init__`.

What you need to do is, in `AWS.__init__`, do `self.aws_instance_type = aws_instance_type` and document `aws_instance_type` in the `Attributes` section of the docstring.

src/dcos_e2e/backends/_aws/__init__.py
```diff
@@ -73,6 +78,8 @@ def __init__(
 
         .. _Regions and Availability Zones:
             https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html
+        .. _Instance types:
```
bamarni: @adamtheturtle : I'm not familiar with this notation, is it correct like this? Or should `..` only be for the first item?

adamtheturtle: This is correct, I believe.

However, to know, build the documentation locally as per http://dcos-e2e.readthedocs.io/en/latest/contributing.html#documentation and see the built document in the local equivalent of http://dcos-e2e.readthedocs.io/en/latest/aws-backend.html#reference.

adamtheturtle:
@bamarni all that remains is this:

> The CI uses Travis CI secrets and so cannot be run from a fork. Are you able to recreate this PR from a branch of dcos/dcos-e2e rather than your own branch?

PR #1037, opened 2018-06-28
Add __eq__ to Node

timaa2k:
This PR adds a `__eq__` function to a `Node` so that `Node`s with the same `public_ip_address` and `private_ip_address` are considered equal when they are compare by `==`.


adamtheturtle:
Great work!

PR #1098, opened 2018-07-09
Execute node-poststart for DC/OS 1.10, 1.11 and master

timaa2k:
This PR adds execution of `node-poststart` to all recent DC/OS versions when calling `Cluster.wait_for_dcos` and `Cluster.wait_for_dcos_ee`.

### DIFF
--- a/src/dcos_e2e/cluster.py
+++ b/src/dcos_e2e/cluster.py
@@ -111,25 +111,34 @@ class Cluster(ContextDecorator):
         tries=500,
         delay=10,
     )
-    def _wait_for_dcos_diagnostics(self) -> None:
+    def _wait_for_node_poststart(self) -> None:
         """
-        Wait until all DC/OS systemd units are healthy.
+        Wait until all DC/OS node-poststart checks are healthy.
+
+        The execution will differ for different version of DC/OS.
+        ``dcos-check-runner`` only exists on DC/OS 1.12+. ``dcos-diagnostics
+        check node-poststart`` only works on DC/OS 1.10 and 1.11. ``3dt`` only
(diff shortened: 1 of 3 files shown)


src/dcos_e2e/cluster.py
```diff
@@ -111,15 +111,28 @@ def from_nodes(
...
         delay=10,
     )
-    def _wait_for_dcos_diagnostics(self) -> None:
+    def _wait_for_node_poststart(self) -> None:
         """
-        Wait until all DC/OS systemd units are healthy.
+        Wait until all DC/OS node-poststart checks are healthy.
+
+        The execution will differ for different version of DC/OS.
+        ``dcos-check-runner`` only exists on DC/OS 1.12+. ``dcos-diagnostics
```
adamtheturtle: ``dcos-diagnostics check node-poststart`` only works on DC/OS 1.10 and 1.11. Is that right? I thought it "works" on 1.12?

Can you add a brief description as per the JIRA about expected flakiness on older versions.

timaa2k: `dcos-diagnostics check node-poststart` functionality is removed from master. There were a couple of commits where both, `dcos-check-runner` and `dcos-diagnostics` had this functionality.
I'll add a comment to the JIRA.

timaa2k: Changing the `dcos-diagnostics --diag` to `dcos-diagnostics check node-poststart` here gets rid of the flakiness on older versions of DC/OS as well since we explicitly execute the checks and wait for them to pass rather than check the state of the systemd unit that executes them when using --diag.

timaa2k:
Merging this as it is a known failing test related to dcos-vagrant that's the issue here.

PR #1123, opened 2018-07-13
Add master host port map for Docker

jieyu:
Add an option to expose host ports for one of the master node. This is
useful on OSX on which the container IP is not directly accessible from
the host. By exposing the host ports, the user can reach the services on
the master node using the mapped host ports. The host port map will be
applied to one master only if there are multiple master nodes.

I tested on my mac locally:
```
$ dcos-docker create --transport docker-exec --public-agents 0 --agents 1 --one-master-host-port-map 80:80 ./dcos_generate_config.ee.sh
default
Cluster "default" has started. Run "dcos-docker wait --cluster-id default" to wait for DC/OS to become ready.
```

### DIFF
--- a/src/cli/dcos_docker/commands/create.py
+++ b/src/cli/dcos_docker/commands/create.py
@@ -78,6 +78,61 @@ def _validate_docker_network(
         raise click.BadParameter(message=message)
 
 
+def _validate_port_map(
+    ctx: click.core.Context,
+    param: Union[click.core.Option, click.core.Parameter],
+    value: Any,
+) -> Dict[str, int]:
+    """
+    Turn port map strings into a Dict that ``docker-py`` can use.
+    """
+    # We "use" variables to satisfy linting tools.
+    for _ in (ctx, param):
+        pass
+
+    ports = {}  # type: Dict[str, int]
+    for ports_definition in value:
+        parts = ports_definition.split(':')
+
+        # Consider support the full docker syntax.
+        # https://docs.docker.com/engine/reference/run/#expose-incoming-ports
+        if len(parts) != 2:
+            message = (
+                '"{ports_definition}" is not a valid port map. '
+                'Please follow this syntax: <HOST_PORT>:<CONTAINER_PORT>'
+            ).format(ports_definition=ports_definition)
+            raise click.BadParameter(message=message)
+
+        host_port, container_port = parts
+        if not host_port.isdigit():
+            message = ('Host port "{host_port}" is not an integer.'
+                       ).format(host_port=host_port)
+            raise click.BadParameter(message=message)
+        if not container_port.isdigit():
+            message = ('Container port "{container_port}" is an integer.'
+                       ).format(container_port=container_port)
+            raise click.BadParameter(message=message)
+        if int(host_port) < 0 or int(host_port) > 65535:
+            message = ('Host port "{host_port}" is not a valid port number.'
+                       ).format(host_port=host_port)
+            raise click.BadParameter(message=message)
+        if int(container_port) < 0 or int(container_port) > 65535:
+            message = (
+                'Container port "{container_port}" is not a valid port number.'
+            ).format(container_port=container_port)
+            raise click.BadParameter(message=message)
+
+        key = container_port + '/tcp'
+        if key in ports:
+            message = (
+                'Container port "{container_port}" specified multiple times.'
+            ).format(container_port=container_port)
+            raise click.BadParameter(message=message)
+
+        ports[key] = int(host_port)
+    return ports
+
+
 def _validate_volumes(
     ctx: click.core.Context,
     param: Union[click.core.Option, click.core.Parameter],
@@ -280,6 +335,17 @@ def _write_key_pair(public_key_path: Path, private_key_path: Path) -> None:
     ),
 )
 @node_transport_option
+@click.option(
+    '--one-master-host-port-map',
+    type=str,
+    callback=_validate_port_map,
+    help=(
+        'Publish a container port of one master node to the host. '
+        'Only Transmission Control Protocol is supported currently. '
+        'The syntax is <HOST_PORT>:<CONTAINER_PORT>'
+    ),
+    multiple=True,
+)
 @click.pass_context
 def create(
     ctx: click.core.Context,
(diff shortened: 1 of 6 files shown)


adamtheturtle:
Sweet! Thanks @jie.

At first glance I'm very positive about this - next is a thorough review.

What I have spotted so far:

* As you said, tests coming next week :smile:
* Please add an entry to `CHANGELOG.rst` under the `Next` heading
* If you think it is appropriate, please add documentation. I guess http://dcos-e2e.readthedocs.io/en/latest/dcos-docker-cli.html#viewing-the-web-ui is a good place. The option will automatically be documented at http://dcos-e2e.readthedocs.io/en/latest/dcos-docker-cli.html#dcos-docker-create.
* The lint build has failed - see http://dcos-e2e.readthedocs.io/en/latest/contributing.html#linting.

I really should make it a thing that validators need test coverage because they seem so easy, but I'll leave that for a boring task sometime for myself!

src/cli/dcos_docker/commands/create.py
```diff
@@ -78,6 +78,56 @@ def _validate_docker_network(
         raise click.BadParameter(message=message)
 
 
+def _validate_port_map(
+    ctx: click.core.Context,
+    param: Union[click.core.Option, click.core.Parameter],
+    value: Any,
+) -> Dict[str, Any]:
```
adamtheturtle: I think that `Any` should be `int`, here and in other places in this PR.

src/dcos_e2e/backends/_docker/_containers.py
```diff
@@ -86,6 +86,7 @@ def start_dcos_container(
     docker_storage_driver: DockerStorageDriver,
     docker_version: DockerVersion,
     network: Optional[docker.models.networks.Network] = None,
+    ports: Dict[str, Any] = {},
```
adamtheturtle: Avoid using dictionary defaults in Python. Dictionaries are mutable, so different function calls can interact, e.g.:

```python
import random

def example(param={}):
	random = random.randint(1, 10)
	param[random] = random
	print(param)

example()
example()
```

can print:

```
{7: 7}
{7: 7, 8: 8}
```

Instead, use:

```python
def foo(
    ports: Optional[Dict[str, Any]] = None,
) -> None:
    ports = ports or {}
```

jieyu: Ah, thanks! Will fix it. My python experience is almost zero :)

src/cli/dcos_docker/commands/create.py
```diff
@@ -78,6 +78,56 @@ def _validate_docker_network(
...
+            message = (
+                '"{hostPort}" is not a valid port number.'
+            ).format(hostPort=hostPort)
+            raise click.BadParameter(message=message)
+        if int(containerPort) < 0 or int(containerPort) > 65535:
+            message = (
+                '"{containerPort}" is not a valid port number.'
+            ).format(containerPort=containerPort)
+
+        ports[containerPort + '/tcp'] = int(hostPort)
```
adamtheturtle: ❓ Do you think that it is worth mentioning in the CLI option help that the protocol is tcp?

src/dcos_e2e/backends/_docker/__init__.py
```diff
@@ -157,6 +158,13 @@ def __init__(
                 custom network on macOS.
                 Therefore, it is recommended that you use this in conjunction
                 with the ``transport`` parameter.
+            one_master_host_port_map: The exposed host ports for one of the
```
adamtheturtle: "one of the master node" > "one of the master nodes".

I'm guessing that "OSX" will cause a spellcheck error - "macOS" will not.

src/cli/dcos_docker/commands/create.py
adamtheturtle: If the user specifies multiple values, and >= 2 have the same container port, there will be no error, but only the last value will be used. I think that there should be an error.

src/cli/dcos_docker/commands/create.py
jieyu: Fixed

src/dcos_e2e/backends/_docker/__init__.py
jieyu: Fixed.

src/cli/dcos_docker/commands/create.py
```diff
@@ -78,6 +78,56 @@ def _validate_docker_network(
...
+            message = (
+                '"{ports_definition}" is not a valid port map. '
+                'Please follow this syntax: <hostPort>:<containerPort>'
+            ).format(ports_definition=ports_definition)
+            raise click.BadParameter(message=message)
+
+        hostPort, containerPort = parts
+        if not hostPort.isdigit():
+            message = (
+                '"{hostPort}" is not a number.'
```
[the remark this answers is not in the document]
jieyu: Fixed.

src/cli/dcos_docker/commands/create.py
jieyu: Done.

adamtheturtle:
@jieyu My apologies! I'm not used to having someone else contribute on the weekend 😄 I merged some changes of mine which pass locally and fail on Travis CI because Travis CI has not got VirtualBox available. Therefore you are seeing unrelated failures. I will fix these now.

That said, all the important tests for your change are passing, I believe.

adamtheturtle:
I have fixed the Vagrant-on-Travis issues.
If you want to see the test suite pass, please merge master in / rebase as you prefer.

Sorry about that!

Ping me when you'd like a review / merge.

tests/test_dcos_e2e/backends/docker/test_docker.py
```diff
@@ -517,3 +517,36 @@ def test_pass_bridge(self) -> None:
...
+            for node in cluster.masters:
+                container = _get_container_from_node(node=node)
+                ports = container.attrs['HostConfig']['PortBindings']
+                if ports is None:
+                    continue
+                if '80/tcp' in ports:
+                    host_ports = ports['80/tcp']
+                    if host_ports[0]['HostPort'] == '8000':
+                        port_map_found = True
+                        break
```
adamtheturtle: ❓ Remove this `break` else you might not hit 100% coverage - the invisible "else" might not be hit if the first node you check is `master-0`.

adamtheturtle:
I'm very happy with this. Please do the following and then we can merge:

* Add a CHANGELOG entry
* Fix the conflicts (sorry)
* Address the minor comment I just added
* Optionally add some documentation

adamtheturtle:
LGTM! Merge when green 😄

PR #1140, opened 2018-07-15
Update Mesos agent cgroup root

### LINKED ISSUE #1124: Cgroup collision between two agents on the same machine using Docker backend.
Tested on my mac. Each agent has 2 cpus, and I ran two private agents.
```sh
$ dcos-docker create --transport docker-exec --public-agents 0 --agents 2 --one-master-host-port-map 80:80 ./dcos_generate_config.ee.sh
```

Then, I deployed a marathon app with 2 instances using the following config:
```json
{
  "id": "/test2",
  "cmd": "sleep 1000",
  "container": {
    "type": "MESOS",
    "docker": {
      "image": "alpine",
    }
  },
  "cpus": 1.5,
  "disk": 0,
  "instances": 2,
  "mem": 128
}
```

Due to the cpu requirements, each agent has one instance running. However, agent_0 can see cgroup created in agent_1 (and vise versa):
```sh
$ dcos-docker run --transport docker-exec ls -ald /sys/fs/cgroup/memory/mesos/*/ --node agent_0
drwxr-xr-x 2 root root 0 Jul 14 05:00 /sys/fs/cgroup/memory/mesos/86cd56b7-0380-4fa3-82e0-8cd540fab7c2/
drwxr-xr-x 2 root root 0 Jul 14 05:00 /sys/fs/cgroup/memory/mesos/e56f1198-e981-4213-a74e-7a59ad857301/
```

```sh
Jies-MacBook-Pro:dcos-e2e jie$ dcos-docker run --transport docker-exec ls -ald /sys/fs/cgroup/memory/mesos/*/ --node agent_1
drwxr-xr-x 2 root root 0 Jul 14 05:00 /sys/fs/cgroup/memory/mesos/86cd56b7-0380-4fa3-82e0-8cd540fab7c2/
drwxr-xr-x 2 root root 0 Jul 14 05:00 /sys/fs/cgroup/memory/mesos/e56f1198-e981-4213-a74e-7a59ad857301/
```

This will cause one agent to kill containers from another agent during restart (i.e., `systemctl restart dcos-mesos-slave`).

jieyu:
Instruct the Mesos agent in the DC/OS Docker container to use a cgroup
root that is under the cgroup used for the DC/OS Docker container.
Otherwise, if one runs multiple DC/OS agent nodes on a single machine,
they might see cgroups from each other, causing issues during recovery
because Mesos agent will kill tasks under unknown cgroups.

Fixes #1124

### DIFF
--- a/src/dcos_e2e/backends/_docker/_containers.py
+++ b/src/dcos_e2e/backends/_docker/_containers.py
@@ -41,7 +41,8 @@ def _docker_service_file(
         '{daemon} '
         '-D '
         '-s {storage_driver_name} '
-        '--exec-opt=native.cgroupdriver=cgroupfs'
+        '--exec-opt=native.cgroupdriver=cgroupfs '
+        '--cgroup-parent=${{CGROUP_PARENT}}'
     ).format(
         storage_driver_name=storage_driver_name,
         daemon=daemon,
@@ -54,6 +55,7 @@ def _docker_service_file(
             'After': 'dbus.service',
         },
         'Service': {
+            'EnvironmentFile': '/etc/docker/env',
             'ExecStart': docker_cmd,
             'LimitNOFILE': '1048576',
             'LimitNPROC': '1048576',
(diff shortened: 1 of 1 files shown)


adamtheturtle:
Thank you @jieyu . Unfortunately this will fail some tests (lint and CLI)  because I broke CI with some Vagrant changes (fixed now). Let's ignore those errors.

Can you confirm that you have manually tested this on macOS?
If so, I am happy to merge when CI is finished.

PR #1161, opened 2018-07-17
Add `master_devices` and `agent_devices` parameters.

nfnt:
These parameters of the Docker backend allow to expose host devices
to master or agent containers.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,11 @@ Changelog
 Next
 ----
 
+2018.07.17.0
+------------
+
+- Add ``master_devices`` and ``agent_devices`` parameters to the ``Docker`` backend.
+
 2018.07.16.0
 ------------
 
--- a/tests/test_dcos_e2e/backends/docker/test_docker.py
+++ b/tests/test_dcos_e2e/backends/docker/test_docker.py
@@ -549,3 +549,61 @@ class TestOneMasterHostPortMap:
                     if host_ports[0]['HostPort'] == '8000':
                         port_map_found = True
             assert port_map_found
+
+
+class TestDevices:
+    """
+    Tests for setting device mapping on master or agent Docker containers.
+    """
+
+    def test_default(self) -> None:
+        with Cluster(
+            cluster_backend=Docker(),
+            masters=1,
+            agents=0,
+            public_agents=0,
+        ) as cluster:
+            (master, ) = cluster.masters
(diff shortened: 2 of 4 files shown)


tests/test_dcos_e2e/backends/docker/test_docker.py
```diff
@@ -549,3 +549,61 @@ def test_one_master_host_port_map(self) -> None:
...
+            masters=1,
+            agents=0,
+            public_agents=0,
+        ) as cluster:
+            (master, ) = cluster.masters
+            container = _get_container_from_node(node=master)
+            devices = container.attrs['HostConfig']['Devices']
+            assert devices is None
+
+    def test_master_devices(self) -> None:
```
adamtheturtle: I think let's have one test which tests all three new parameters - that way, we can say "`master_devices` does not interfere with agents".

CHANGELOG.rst
```diff
@@ -6,6 +6,11 @@ Changelog
 Next
 ----
 
+2018.07.17.0
+------------
+
+- Add ``master_devices`` and ``agent_devices`` parameters to the ``Docker`` backend.
```
adamtheturtle: Let's add a separate `public_agent_devices`.

jieyu:
I don't know if this is useful for the storage testing case. This just exposes a devices from host to the container. But how do you know about those devices on the host? 

I was hoping that I don't need to do any work on the host. Instead, we should setup some loopback devices in the docker container (the number of devices can be configured) so that `dcos-docker create` is sufficient for testing.

nfnt:
@jieyu Good point! I was thinking of creating a loopback device on the host and expose it in the container, but you're right, that it would be much better to setup the loopback device directly in the container. This should work with how containers are currently setup, @adamtheturtle? Then we would have to run a `dd ... && losetup ...` on a running container based on some config value.

jieyu:
@nfnt I am aware this issue. And we do have a few workarounds in our e2e container to solve this:
https://github.com/mesosphere/dcos-storage/blob/master/scripts/docker/e2e/loopback/setup_loopback_volumes.sh

We can smartly pick an unused loopback device.

From user's perspective, asking them to setup loopback is too tedious...

nfnt:
I'm closing this in favor of #1285, which creates a sidecar container to use for the loopback device. This method works in the API, as well as the `dcos-docker` CLI without having the user to create any devices on their host.

PR #1167, opened 2018-07-21
Remove branching in a test in an attempt to get to 100% test branch c…

adamtheturtle:
…overage


adamtheturtle:
This has improved coverage but shows that recent changes in DC/OS Enterprise have reduced coverage - I will update that in another PR.

PR #1177, opened 2018-07-23
Load ip_detect_contents from ip-detect file for each backend

timaa2k:


adamtheturtle:
Nice work @timaa2k ! 

Merge when you're happy.

PR #1199, opened 2018-07-24
Move ip_detect_path from Cluster into the ClusterBackend

timaa2k:


adamtheturtle:
Thanks @timaa2k - merge when you are happy.

PR #1203, opened 2018-07-24
Add functionality of copying directories onto a Node via Node.send_file

timaa2k:
https://jira.mesosphere.com/browse/DCOS_OSS-3859

### DIFF
--- a/src/dcos_e2e/node.py
+++ b/src/dcos_e2e/node.py
@@ -551,7 +551,8 @@ class Node:
             tar.add(str(local_path), arcname=str(arcname), recursive=True)
 
         # `remote_path` may be a tmpfs mount.
-        # At the time of writing, for example, `/tmp` is a tmpfs mount.
+        # At the time of writing, for example, `/tmp` is a tmpfs mount
+        # on the Docker backend.
         # Copying files to tmpfs mounts fails silently.
         # See https://github.com/moby/moby/issues/22020.
         home_path = self.run(
@@ -560,6 +561,8 @@ class Node:
             transport=transport,
             sudo=True,
         ).stdout.strip().decode()
+        # Therefore, we create a temporary file within our home directory,
+        # We then remove the temporary file at the end of this function.
 
         remote_tar_path = Path(home_path) / tar_name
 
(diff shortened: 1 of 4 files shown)


src/dcos_e2e/node.py
```diff
@@ -536,21 +537,67 @@ def send_file(
...
+        tempdir = Path(gettempdir())
+        tar_name = '{unique}.tar'.format(unique=uuid.uuid4().hex)
+        local_tar_path = tempdir / tar_name
+
+        with tarfile.open(str(local_tar_path), 'w') as tar:
+            arcname = remote_path.relative_to(remote_path.parent)
+            tar.add(str(local_path), arcname=str(arcname), recursive=True)
+
+        # `remote_path` may be a tmpfs mount.
+        # At the time of writing, for example, `/tmp` is a tmpfs mount.
```
adamtheturtle: ...on the Docker backend

adamtheturtle: I think copy the rest of the comment from the `docker_exec` transport:

```
        # Therefore, we create a temporary directory within our home directory,
        # and we put the file there.
        # We then move the file from the temporary directory to the intended
        # destination.
        # We then remove the temporary directory.
```

adamtheturtle:
Thanks @timaa2k ! Merge when you are happy or ping me if there are changes beyond the suggested comment changes.

tests/test_dcos_e2e/test_node.py
```diff
@@ -158,6 +158,33 @@ def test_send_file(
...
+        local_file = local_dir.join(file_name)
+        local_file.write(content)
+        random = uuid.uuid4().hex
+        master_base_dir = '/etc/{random}'.format(random=random)
+        master_destination_dir = Path(master_base_dir) / remote_dir_name
+        dcos_node.send_file(
+            local_path=Path(local_dir),
+            remote_path=master_destination_dir,
+        )
+        args = ['cat', str(master_destination_dir / file_name)]
```
adamtheturtle: Ok so to clarify, this puts the contents of the local dir into the remote dir - right?

Let's go with whatever is easiest for now, and change as we see how this interface plays out.

adamtheturtle: However, and I know we discussed this, this isn't how I'd expect it to work seeing it now.

I'd expect this to be like `cp -r`.

PR #1235, opened 2018-07-26
SELinux enforce option on the dcos-aws CLI

timaa2k:


src/cli/dcos_aws/commands/create.py
```diff
@@ -76,6 +76,14 @@
 @copy_to_master_option
 @verbosity_option
 @cluster_id_option
+@click.option(
```
adamtheturtle: Please make this into a common option with `cli.common.options`.

adamtheturtle:
Thanks @timaa2k ! Please address the minor comment, test manually and merge.

adamtheturtle:
`test_dcos_aws/test_cli` tests will fail. Please update the help text there.

PR #1285, opened 2018-08-20
Block device support using loopback devices.

nfnt:
This makes it possible to provide an unformatted block device to a cluster
that is using the Docker backend.
Support for this is added to the API (via `DockerLoopbackVolume`) as well as
the `dcos-docker` CLI (via the `loopback-device` parameter). It creates
a sidecar container containing an unformatted file that is mounted to all
nodes as a loopback device.


src/cli/dcos_docker/commands/create.py
```diff
@@ -195,6 +196,20 @@ def _validate_volumes(
     return mounts
 
 
+def _validate_loopback_device(
```
adamtheturtle: Instead use `type=click.IntRange` on the option.

src/dcos_e2e/backends/_docker/__init__.py
```diff
@@ -96,6 +96,124 @@ def _get_fallback_storage_driver() -> DockerStorageDriver:
...
+        Args:
+            size: Size of the block device in Megabytes.
+            labels: Docker labels to add to the container.
+        Returns:
+            A tuple containing the container and the path of
+            the loopback device.
+        """
+        client = docker.from_env(version='auto')
+
+        container = client.containers.create(
```
bbannier: If `centos:latest` is not already on the host, this `create` would fail; We need to pull it here to make sure.

src/cli/dcos_docker/commands/create.py
nfnt: Wouldn't I have to provide a max value in that case? Hm, probably could say something like `int.MAX` here.

adamtheturtle: I believe `max=None` will work.

src/dcos_e2e/backends/_docker/__init__.py
adamtheturtle: I think `create` does create and pull - but please check.

src/dcos_e2e/backends/_docker/__init__.py
```diff
@@ -96,6 +96,124 @@ def _get_fallback_storage_driver() -> DockerStorageDriver:
...
+            privileged=True,
+            detach=True,
+            tty=True,
+            image='centos',
```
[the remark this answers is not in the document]
nfnt: To pull the same image that will be used by DC/OS. Ideally the layers are already downloaded for building DC/OS. Though it doesn't look like that at the moment. Will look into this.

src/cli/dcos_docker/commands/_common.py
```diff
@@ -44,6 +46,18 @@
...
+
+class Role(Enum):
+    """
+    Roles of DC/OS nodes.
+    """
+
+    MASTER = 'master'
+    AGENT = 'slave'
+    PUBLIC_AGENT = 'slave_public'
+    LOOPBACK_SIDECAR = 'loopback'
```
adamtheturtle: This doesn't seem like a it fits with " Roles of DC/OS nodes.".

src/dcos_e2e/backends/_docker/__init__.py
```diff
@@ -96,6 +96,124 @@ def _get_fallback_storage_driver() -> DockerStorageDriver:
         return DockerStorageDriver.AUFS
 
 
+class DockerLoopbackVolume():
```
adamtheturtle: I'm a little torn because, ignoring the CLI I don't think that this belongs as part of DC/OS E2E's library - it doesn't really fit in my concept of what E2E is for.

However, I do see value in having this in the CLI, as you have explained to me that this would be useful as part of the CLI.

Ideally for me this would be a separate library in itself, but that is probably not going to happen.

Therefore, let's discuss where this goes - is `backends.DockerLoopbackVolume` natural? I think that this implies that it is a backend which it is not.

src/dcos_e2e/backends/_docker/__init__.py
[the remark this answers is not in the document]
adamtheturtle: Ok this is not quite a correct assumption: you cannot rely on which image(s) will be used for DC/OS.

At the time of writing the following images are used for DC/OS:

```
ubuntu:xenial
centos:7
quay.io/shift/coreos:stable-1298.7.0
```

Please use tags in the images to increase stability.

tests/test_dcos_e2e/backends/docker/test_docker.py
```diff
@@ -556,3 +556,24 @@ def test_one_master_host_port_map(self) -> None:
...
             }
             assert master_port_settings == expected_master_port_settings
+
+
+class TestDevices:
+    """
+    Tests for setting device mapping on master or agent Docker containers.
+    """
+
+    def test_loopback(self) -> None:
```
adamtheturtle: Please add a docstring which describes what is true, given this test passing.

src/cli/dcos_docker/commands/create.py
```diff
@@ -456,6 +483,16 @@ def create(
     )
 
     try:
+        if loopback_device is not None:
```
adamtheturtle: This block will not raise a `CalledProcessError`, right? If so, it should be outside of the `try` block.

tests/test_dcos_e2e/backends/docker/test_docker.py
```diff
@@ -556,3 +556,24 @@ def test_one_master_host_port_map(self) -> None:
                 }],
             }
             assert master_port_settings == expected_master_port_settings
+
+
+class TestDevices:
```
adamtheturtle: Please add a test for each of the public methods in `DockerLoopbackVolume`.

Please add a test for the `labels` parameter.

tests/test_dcos_e2e/backends/docker/test_docker.py
```diff
@@ -556,3 +556,24 @@ def test_one_master_host_port_map(self) -> None:
...
+
+    def test_loopback(self) -> None:
+        with DockerLoopbackVolume(size=256) as device:
+            with Cluster(
+                cluster_backend=Docker(),
+                masters=1,
+                agents=1,
+                public_agents=0,
+            ) as cluster:
+                (agent, ) = cluster.agents
```
adamtheturtle: It is not clear why an agent is tested and not a master.

src/cli/dcos_docker/commands/_common.py
```diff
@@ -194,6 +209,13 @@ def public_agents(self) -> Set[Container]:
         """
         return self._containers_by_role(role=Role.PUBLIC_AGENT)
 
+    @property
+    def loopback_sidecars(self) -> Set[Container]:
```
adamtheturtle: It looks like this can be private.

src/cli/dcos_docker/commands/create.py
```diff
@@ -315,6 +330,17 @@ def _validate_volumes(
...
 )
+@click.option(
+    '--loopback-device',
+    type=int,
+    callback=_validate_loopback_device,
+    help=(
+        'Create a block device mounted as loopback device that will be '
+        'available on all nodes. The value indicates the size in '
+        'megabytes.\n'
+        'WARNING: Make sure that only a single node is using this device!'
```
adamtheturtle: What happens if multiple nodes use a particular device?

src/cli/dcos_docker/commands/create.py
```diff
@@ -315,6 +330,17 @@ def _validate_volumes(
     ),
     multiple=True,
 )
+@click.option(
```
adamtheturtle: Might a user want multiple of these?

adamtheturtle:
Loopback devices, like custom Docker networks, can apply to multiple clusters.
Therefore, I do not think that the loopback device option should be part of `dcos-docker create`.

This seems as separate to DC/OS E2E as creating a Docker network.
To use a custom Docker network, one does the following:

```
$ docker network create custom-bridge
$ dcos-docker create /path/to/dcos_generate_config.sh
    --network custom-bridge
    --genconf-dir ./custom-genconf
```

I therefore imagine using a loopback-device as this:

```
$ some-loopback-cli create --size 50mb --name example
$ dcos-docker create /path/to/dcos_generate_config.sh
$ some-loopback-cli destroy --name example
```

To me this isn't necessarily part of DC/OS E2E.
It could be a separate project / repo entirely.
That said, we do have utilities in DC/OS E2E which help the use of DC/OS E2E.
For example, we have `dcos-docker setup-mac-network`.

So, if I were to imagine this as part of DC/OS E2E, it would look like:

```
$ dcos-docker create-loopback-sidecar --size 50mb --name example
$ dcos-docker create /path/to/dcos_generate_config.sh
$ dcos-docker destroy-loopback-sidecar --name example
```

(or `$ dcos-docker loopback create --size 50mb --name example`).

And as part of the library, the context manager might live in `dcos_e2e.docker_utils`.

What do you think @nfnt ?

src/dcos_e2e/backends/_docker/__init__.py
adamtheturtle: Please document this in the public documentation (in `docs/source`).

nfnt:
I like the idea of having separate commands for creating/destroying these sidecar containers. They would provide a platform-independent way to handle this problem while at the same time avoiding accidental leaks by shifting the responsibility to destroy the container to the user.
But I'd very much have this as part of DC/OS E2E, as subcommands of `dcos-docker`.

jieyu:
I got confused. What's the process (i.e., commands) to launch 2 DC/OS agents using docker, each has one loopback device available to be used by DC/OS?

nfnt:
@jieyu I don't understand the question. Every loopback device is visible to every node in a cluster -- well, even to every running container. One can create multiple loopback sidecars, but each sidecar has to be accessed by a single node at a time.

src/dcos_e2e/docker_utils/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+            the loopback device.
+        """
+        client = docker.from_env(version='auto')
+
+        container = client.containers.create(
+            name='dcos-e2e-loopback-sidecar',
+            privileged=True,
+            detach=True,
+            tty=True,
+            image='centos:7',
```
adamtheturtle: Not complete thought:

IIRC we discussed: Does this pull the image? Why CentOS 7? Is this pulled by `make pull-images` (so the network access in CI is minimised during test run time)?

src/dcos_e2e/docker_utils/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+            size: Size of the block device in Megabytes.
+            labels: Docker labels to add to the container.
+        """
+        self.size = size
+        docker_container_labels = labels or {}
+
+        self.container, self.path = \
+            DockerLoopbackVolume.create(size, docker_container_labels)
+
+    @staticmethod
```
adamtheturtle: Why use `staticmethod` for`create` and `destroy`?

adamtheturtle:
Please add commands to `make clean` in `Makefile` so that loopback sidecars are cleaned up.

tests/test_dcos_e2e/docker_utils/test_loopback.py
```diff
@@ -0,0 +1,69 @@
...
+            container.remove(v=True)
+
+    def test_destroy(
+        self,
+        loopback_sidecar: Tuple[docker.models.containers.Container, str],
+    ) -> None:
+        """
+        Calling `DockerLoopbackVolume.destroy` destroys a sidecar container.
+        """
+        DockerLoopbackVolume.destroy(loopback_sidecar[0])
```
adamtheturtle: As a reader of this test it is not clear to me how this demonstrates what the test docstring says.

docs/source/dcos-docker-cli.rst
```diff
@@ -369,3 +369,14 @@ CLI Reference
 
 .. click:: cli.dcos_docker:download_artifact
   :prog: dcos-docker download-artifact
+
```
adamtheturtle: ❓ What do you think of adding a section in this document on using loopback sidecars?

src/cli/dcos_docker/commands/destroy_loopback_sidecar.py
```diff
@@ -0,0 +1,59 @@
...
+                value=NODE_TYPE_LOOPBACK_SIDECAR_LABEL_VALUE,
+            ),
+            '${key}=${value}'.format(
+                key=SIDECAR_NAME_LABEL_KEY,
+                value=name,
+            ),
+        ],
+    }
+    containers = client.containers.list(filters=filters)
+    assert len(containers) <= 1
```
adamtheturtle: Will it be clear to a user what the problem is if this `AssertionError` is raised?

docs/source/dcos-docker-cli.rst
adamtheturtle: For example noting that one sidecar is used by all containers.
This even confused @jieyu I think.

src/dcos_e2e/docker_utils/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+            cmd=['/bin/bash', '-c', write_device_path],
+        )
+        assert exit_code == 0, output.decode()
+
+        return (container, path)
+
+    @staticmethod
+    def destroy(container: docker.models.containers.Container) -> None:
+        """
+        Destroy a container containing a block device that is used with
```
adamtheturtle: The function does not seem to destroy a container like the docstring says.

tests/test_dcos_e2e/docker_utils/test_loopback.py
```diff
@@ -0,0 +1,69 @@
...
+
+        with DockerLoopbackVolume(size=1, labels=labels) as device:
+            cmd = 'lsblk {}'.format(device.path)
+            exit_code, output = device.container.exec_run(cmd=cmd)
+            assert exit_code == 0, cmd + ': ' + output.decode()
+
+            for key, value in labels.items():
+                assert device.container.labels[key] == value
+
+    def test_create(self) -> None:
```
adamtheturtle: What happens if I run `create` twice? Is this supported?

adamtheturtle:
Please add documentation for the new class.

tests/test_dcos_e2e/docker_utils/test_loopback.py
```diff
@@ -0,0 +1,69 @@
...
+        """
+        Calling `DockerLoopbackVolume.create` creates a sidecar container.
+        """
+        labels = {'foo': 'bar'}
+
+        container, path = DockerLoopbackVolume.create(
```
adamtheturtle: What do you think about:

```
loopback_volume = DockerLoopbackVolume(size=1, labels=labels)
container = loopback_volume.container
path = loopback_volume.path
```

?

src/dcos_e2e/docker_utils/__init__.py
nfnt: They're needed for the CLI. There we can't  use an instance of `DockerLoopbackVolume` when destroying an existing sidecar, because it would create a new sidecar during instantiation.
An alternative would be to have multiple constructors (or a factory) for `DockerLoopbackVolume`, allowing us to create an instance from an existing `docker.models.containers.Container`.

src/cli/dcos_docker/commands/destroy_loopback_sidecar.py
nfnt: I'll add an error message here.

docs/source/dcos-docker-cli.rst
nfnt: Adding additional documentation.

src/dcos_e2e/docker_utils/__init__.py
nfnt: Thanks, I'll update the function to include `container.stop` and `container.remove` calls.

tests/test_dcos_e2e/docker_utils/test_loopback.py
nfnt: Good point! I'll add support for that.

tests/test_dcos_e2e/docker_utils/test_loopback.py
nfnt: I'm explictly testing the static `DockerLoopbackVolume.create` method here.

docs/source/dcos-docker-cli.rst
```diff
@@ -302,6 +302,25 @@ It is possible to use :ref:`dcos-docker-create` to create a cluster with a custo
 
    See `Verify installation <https://docs.mesosphere.com/1.11/security/ent/tls-ssl/ca-custom/#verify-installation>`_ for steps to verify that the DC/OS Enterprise cluster was installed properly with the custom CA certificate.
 
+Using a loopback sidecar
+------------------------
+
+The :ref:`dcos-docker-create-loopback-sidecar` command can be used to create a
+loopback sidecar. This will provide all containers with a unformatted block
+device, mounted as a loopback device. It can be destroyed with
+:ref:`dcos-docker-destroy-loopback-sidecar`. All container have access to this
```
adamtheturtle: `s/container/containers`.

src/cli/dcos_docker/commands/destroy_loopback_sidecar.py
```diff
@@ -0,0 +1,58 @@
...
+                value=NODE_TYPE_LOOPBACK_SIDECAR_LABEL_VALUE,
+            ),
+            '{key}={value}'.format(
+                key=SIDECAR_NAME_LABEL_KEY,
+                value=name,
+            ),
+        ],
+    }
+    containers = client.containers.list(filters=filters)
+    assert len(containers) <= 1, \
```
adamtheturtle: Instead, let's validate that the name is unique when creating the sidecar.

tests/test_cli/test_dcos_docker/test_cli.py
```diff
@@ -1347,3 +1349,90 @@ def test_help(self) -> None:
...
+            Options:
+              --help  Show this message and exit.
+            """,# noqa: E501,E261
+        )
+        # yapf: enable
+        assert result.output == expected_help
+
+    def test_sidecar_container_does_not_exist(self) -> None:
+        """
+        An error is shown if the given cluster does not exist.
```
adamtheturtle: `s/cluster/sidecar container`.

tests/test_dcos_e2e/docker_utils/test_loopback.py
adamtheturtle: @nfnt Any chance you can address this?

nfnt: I've updated the docstring to clarify that in case of a failure an expection would be thrown.

src/dcos_e2e/docker_utils/__init__.py
nfnt: Added a comment on why CentOS 7 is chosen here.

adamtheturtle:
AWESOME WORK @nfnt !!! ❤️  Merging

src/dcos_e2e/docker_utils/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+        create_loopback_device = (
+            'dd if=/dev/zero of=/volume0 bs=1M count={size};'
+        ).format(size=size)
+
+        exit_code, output = container.exec_run(
+            cmd=['/bin/bash', '-c', create_loopback_device],
+        )
+        assert exit_code == 0, output.decode()
+
+        setup_loopback_device = 'losetup --find --show /volume0;'
```
jdef: Future work: `losetup` is unreliable due to race conditions in the lo driver itself. I don't think this has yet been patched yet on the RHEL/CentOS kernels. If the Docker's host OS has a recent enough kernel then this won't be a problem. This is not the case, for example, w/ CI testing for the dcos-storage project. Hence, dcos-storage "in-repo" test scripts jump through a few hoops to "validate" the loopback device before trying to use it - this avoids flakes both in local development and on CI.

PR #1288, opened 2018-08-21
Update requirements, setting new pyroma minimum

adamtheturtle:


adamtheturtle:
https://github.com/sphinx-doc/sphinx/issues/5322 - we can't update Sphinx just yet because of this.

PR #1322, opened 2018-09-25
Add a timeout and a custom DCOSTimeoutError to wait_for_dcos_oss/ee

timaa2k:
In order to no indefinitely wait for DC/OS to boot up, we need to limit the time spent waiting for all the components to come up.

### DIFF
--- a/docs/source/exceptions.rst
+++ b/docs/source/exceptions.rst
@@ -1,6 +1,6 @@
 Exceptions
 ==========
 
-The following custom exceptions are defined dcos-e2e.
+The following custom exceptions are defined in dcos-e2e.
 
 .. autoclass:: dcos_e2e.exceptions.DCOSTimeoutError
--- a/docs/source/exceptions.rst
+++ b/docs/source/exceptions.rst
@@ -1,0 +1,6 @@
+Exceptions
+==========
+
+The following custom exceptions are defined dcos-e2e.
+
+.. autoclass:: dcos_e2e.exceptions.DCOSTimeoutError
--- a/src/dcos_e2e/cluster.py
+++ b/src/dcos_e2e/cluster.py
@@ -171,9 +171,14 @@ class Cluster(ContextDecorator):
                 start up is considered a failure.
 
         Raises:
+            ValueError: In case ``timeout_seconds`` is < 1.
             dcos_e2e.exceptions.DCOSTimeoutError: Raised if cluster components
-                did not become ready within timeout the boundary.
+                did not become ready within the timeout boundary.
         """
+
+        if timeout_seconds < 1:
+            message = '``timeout_seconds`` must be ``None`` or >= 1.'
+            raise ValueError(message)
 
         @timeout_decorator.timeout(
             timeout_seconds,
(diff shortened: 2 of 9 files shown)


adamtheturtle:
This change needs a `CHANGELOG` entry.

src/dcos_e2e/cluster.py
```diff
@@ -117,7 +120,6 @@ def from_nodes(
 
     @retry(
         exceptions=(subprocess.CalledProcessError),
-        tries=500,
```
adamtheturtle: This is a behavioural change. Before - this timed out by default. Now, this does not time out unless the parent function is given a `timeout` value.

timaa2k: Do you think this is undesirable? I think it adds consistency.

src/dcos_e2e/cluster.py
```diff
@@ -151,7 +153,11 @@ def _wait_for_node_poststart(self) -> None:
                 shell=True,
             )
 
-    def wait_for_dcos_oss(self, http_checks: bool = True) -> None:
+    def wait_for_dcos_oss(
+        self,
+        http_checks: bool = True,
+        timeout: int = 0,
```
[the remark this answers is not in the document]
timaa2k: I thought about None as well but it's gonna ugly to pass this to the decorator -> need to have two functions, one decorated.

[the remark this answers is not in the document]
adamtheturtle: Better we take the hit of ugliness than all users taking the hit of a poor interface. Can we not at the top of this function say:

```python
if timeout is None:
    timeout = 0
```

?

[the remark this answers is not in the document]
timaa2k: What happens if you pass in 0 then? :)

src/dcos_e2e/cluster.py
```diff
@@ -269,74 +301,95 @@ def wait_for_dcos_ee(
...
 
-        enterprise_session.wait_for_dcos()
+            if ssl_enabled:
+                try:
+                    response = enterprise_session.get(
+                        # We wait for 10 minutes which is arbitrary but should
+                        # be more than enough after all systemd units are
+                        # healthy.
+                        '/ca/dcos-ca.crt',
+                        retry_timeout=60 * 10,
```
adamtheturtle: Can we just have this be an infinite timeout then?

adamtheturtle: Then we don't need to catch the exception

adamtheturtle: To be clear - as written, this means that our users can give `None` and still get a `DCOSTimeoutError`.

src/dcos_e2e/cluster.py
```diff
@@ -269,74 +301,95 @@ def wait_for_dcos_ee(
...
+                    )
+                except retrying.RetryError:  # pragma: no cover
+                    raise DCOSTimeoutError
+
+                response.raise_for_status()
+                enterprise_session.set_ca_cert()
+
+            try:
+                enterprise_session.wait_for_dcos()
+            except retrying.RetryError:  # pragma: no cover
```
adamtheturtle: This means that our users can give `None` and still get a `DCOSTimeoutError`.

docs/source/exceptions.rst
```diff
@@ -0,0 +1,6 @@
+Exceptions
+==========
+
+The following custom exceptions are defined dcos-e2e.
```
adamtheturtle: I guess you mean "defined *in* dcos-e2e". Either way let's just do "defined." I think.

src/dcos_e2e/cluster.py
```diff
@@ -161,102 +167,128 @@ def wait_for_dcos_oss(self, http_checks: bool = True) -> None:
...
                 cannot be made to the cluster. For example, this is useful on
                 macOS without a VPN set up.
+            timeout: Timeout after which a DC/OS start up is considered to be
+                a failure. The value 0 leads to an infinite timeout.
 
         Raises:
-            RetryError: Raised if any cluster component did not become
-                healthy in time.
+            dcos_e2e.exceptions.DCOSTimeoutError: Raised if cluster components
+                did not become ready within timeout the boundary.
```
adamtheturtle: "within timeout the boundary" - this is not grammatically correct.

PR #1324, opened 2018-09-28
Add test that streaming command output works for any transport

timaa2k:


adamtheturtle:
Approve modulo CI.

Feel free to cancel all builders except the one running the new test, and the linter, and then merge.

adamtheturtle:
From discussion with @timaa2k , this is not a useful test.

PR #1546, opened 2018-11-23
Get ready for 1.13 IAM changes to the IAM

adamtheturtle:


orsenthil:
Thanks, @adamtheturtle  - the using api for authentication reads much better!

PR #1548, opened 2018-11-26
Test that created loopback devices have expected size.

nfnt:


tests/test_dcos_e2e/docker_utils/test_loopback.py
```diff
@@ -18,21 +18,34 @@ def test_loopback(self) -> None:
...
+                '-eq',
+                '"1048576"',
+                ']',
+            ]
+            exit_code, output = new_container.exec_run(
+                cmd=block_device_has_right_size,
+            )
             assert exit_code == 0, device.path + ': ' + output.decode()
 
+            new_container.stop()
```
adamtheturtle: Let's move these back above the assertion. That means that even if the assertion fails, the container will be removed.

adamtheturtle:
Thanks @nfnt - as CockroachDB was shipped on Friday to DC/OS OSS, our CI is broken, so let's be happy as long as this test and the linter pass.

tests/test_dcos_e2e/docker_utils/test_loopback.py
```diff
@@ -18,21 +18,34 @@ def test_loopback(self) -> None:
         client = docker.from_env(version='auto')
 
         with DockerLoopbackVolume(size=1) as device:
```
adamtheturtle: This makes me think - what do you think of changing this parameter name to be something like `size_megabytes`?

tests/test_dcos_e2e/docker_utils/test_loopback.py
nfnt: Good point! Though I'd rather use a `try... finally` block here, as there's more than one assertion that could fail.

tests/test_dcos_e2e/docker_utils/test_loopback.py
nfnt: The best solution would be a `Size` type, but I'm coming from statically typed languages, not sure if this is a common pattern in Python.

adamtheturtle: The most common thing is to use something like `size_megabytes`. However, I have used [bitmath](https://pypi.org/project/bitmath/) to provide a type for sizes.

adamtheturtle: I'll leave it to you to choose what to do, where doing nothing is an acceptable (to me) but not preferred choice. In a follow-up I hope so we can ship this.

tests/test_dcos_e2e/docker_utils/test_loopback.py
adamtheturtle: Either works for me.

tests/test_dcos_e2e/docker_utils/test_loopback.py
```diff
@@ -18,21 +18,34 @@ def test_loopback(self) -> None:
...
+
+            block_device_exists_cmd = ['lsblk', device.path]
             exit_code, output = new_container.exec_run(
                 cmd=block_device_exists_cmd,
             )
-            new_container.stop()
-            new_container.remove()
+            assert exit_code == 0, device.path + ': ' + output.decode()
 
+            block_device_has_right_size = [
```
adamtheturtle: Optional change:

Instead use a command which prints the device size, then make the comparison of something like `assert output == '1048576'`. That will help us find out what the size actually is if this comparison fails.

adamtheturtle:
Thanks @nfnt ! I'm happy. Please make the change that means we are less likely to have leftover containers if something fails, then choose whether to make the other suggested changes / ignore them / do them in a separate PR.

tests/test_dcos_e2e/docker_utils/test_loopback.py
nfnt: `size_megabytes` it is. Will create an extra patch for that.

adamtheturtle:
Thanks @nfnt this fixes https://jira.mesosphere.com/browse/DCOS_OSS-4359.

PR #1586, opened 2019-01-16
Fix exception handling in `DockerLoopbackVolume`

adamtheturtle:
cc @nfnt 

Our nightly builder showed that we had missing coverage in the changed test.
What was happening was that an ``AssertionError`` was being raised, but our `__exit__` had `return True` which caught exceptions.

When this was changed, I saw various errors, in order:

* The container did not exist when the `exec` commands were run - I made a `sleep infinity` entrypoint
* The `stat` command we had used `-f` for format - but this should be `-c --format`
* The `stat` command format was not what we wanted, so I chose to use `blockdev`

I since reformatted the test to remove the `try... finally` block.

### DIFF
--- a/tests/test_dcos_e2e/docker_utils/test_loopback.py
+++ b/tests/test_dcos_e2e/docker_utils/test_loopback.py
@@ -21,7 +21,7 @@ class TestDockerLoopbackVolume:
             privileged=True,
             detach=True,
             image='centos:7',
-            entrypoint =['/bin/sleep', 'infinity'],
+            entrypoint=['/bin/sleep', 'infinity'],
         )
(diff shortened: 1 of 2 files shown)


tests/test_dcos_e2e/docker_utils/test_loopback.py
```diff
@@ -11,41 +11,37 @@ class TestDockerLoopbackVolume:
     Tests for setting device mapping on master or agent Docker containers.
     """
 
-    def test_loopback(self) -> None:
+    @pytest.mark.parametrize('size_megabytes', [1, 2])
```
adamtheturtle: I also basically added a test that the size wasn't hardcoded to 1

PR #1730, opened 2019-04-19
Remove logging handler config from e2e library

mhrabovcin:


adamtheturtle:
Thanks @mhrabovcin ! Looking at the test result, I can see the following:

* A lint error from `master` - sorry, I fixed this in `master` now!
* A CLI test failure from `master`  - sorry, I fixed this in `master` now!
* A genuine test failure in `test_log_and_capture`:

```
    @pytest.mark.parametrize(
        'stdout_message',
        [uuid.uuid4().hex, 'å'],
        ids=['ascii', 'unicode'],
    )
    @pytest.mark.parametrize(
        'stderr_message',
        [uuid.uuid4().hex, 'å'],
        ids=['ascii', 'unicode'],
    )
    def test_log_and_capture(
        self,
        caplog: LogCaptureFixture,
        dcos_node: Node,
        stdout_message: str,
        stderr_message: str,
    ) -> None:
        """
        When given ``Output.LOG_AND_CAPTURE``, stderr and stdout are captured
        in the output as stdout.
    
        stdout and stderr are logged.
        """
        args = ['echo', stdout_message, '&&', '>&2', 'echo', stderr_message]
        result = dcos_node.run(
            args=args,
            shell=True,
            output=Output.LOG_AND_CAPTURE,
        )
    
        # stderr is merged into stdout.
        # This is not ideal but for now it is the case.
        # The order is not necessarily preserved.
        expected_messages = set([stdout_message, stderr_message])
        result_stdout = result.stdout.strip().decode()
        assert set(result_stdout.split('\n')) == expected_messages
    
>       first_log, second_log = caplog.records
E       ValueError: not enough values to unpack (expected 2, got 0)
```

Please can you address that test failure.

src/dcos_e2e/_common.py
```diff
@@ -7,40 +7,7 @@
...
-        name: Name of the logger. Setting this to ``__name__`` will log
-            the path to the file where the logger has been created.
-
-    Returns:
-        See :py:class:`logging.Logger`.
-    """
-    # This gets the root logger with a ``logging.lastResort``
-    # StreamHandler that logs on WARNING level.
-    logger = logging.getLogger(name=name)
-    logger.setLevel(logging.DEBUG)
```
adamtheturtle: Will a test suite which uses DC/OS E2E now have to do something in order to see useful information - something it did not have to do before?

If so, can that go in the CHANGELOG please.

adamtheturtle:
Nice to see `_common` code going away! I want to get rid of this file after this change or rename it to `_subprocess_tools`!

adamtheturtle:
Thanks @mhrabovcin - this looks great to me.
Thank you for working on something a little tedious which will make a big difference.

Let's of course be aware that we might learn that there is a great, still valid, reason behind the old code but we can come to that if and when we spot any errors.

I'm happy to merge (or for you to merge) when CI passes modulo known flakes.

FYI flakes I know of:

* Downloading the artefact on a node has no retry and can error
* Weird "No output received" type errors from Travis CI early on in the build process

Also, Travis CI errors when there has not been 10 minutes of output from a test. If this change means that some tests are silent for extended periods of time then we may have to change something.

adamtheturtle:
Nice work! I'll merge and do a release.

PR #1754, opened 2019-05-03
Use correct ip address in ClusterContainers to_node

brunodebus:
When creating a minidcos docker cluster with a custom docker network the
ip address of the docker0 network is used and not the one of the custom
docker network. Looks like this got introduced by
540b4e16461d27109cc7d8ed37d31f8eecd86d64 (before that it worked just
fine). Reuse the code from DockerCluster _nodes to fix this.

### DIFF
--- a/src/dcos_e2e_cli/dcos_docker/commands/_common.py
+++ b/src/dcos_e2e_cli/dcos_docker/commands/_common.py
@@ -101,7 +101,11 @@ class ClusterContainers(ClusterRepresentation):
         Return the ``Node`` that is represented by a given ``container``.
         """
         container = node_representation
-        address = IPv4Address(container.attrs['NetworkSettings']['IPAddress'])
+        networks = container.attrs['NetworkSettings']['Networks']
+        network_name = 'bridge'
+        if len(networks) != 1:
+            [network_name] = list(networks.keys() - set(['bridge']))
+        address = IPv4Address(networks[network_name]['IPAddress'])
         return Node(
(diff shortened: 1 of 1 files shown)


adamtheturtle:
Really nice work @brunodebus - and thank you for digging through the frankly very messy history!

This looks good.
I will try this manually and then if it works I will do a release ASAP.

adamtheturtle:
@brunodebus Can you please specify a workflow which does not work on the latest release / master but does work with this change?

brunodebus:
I typically create a new docker bridge per platform. (I'm doing this to make sure no other containers run on the same network, which in turn allows me to stop and revive the docker images making up the complete cluster while keeping the ip addresses the same => useful if I turn off my pc or want to try something on a new platform while preserving the old state of another platform).

So my flow is pretty simple:
```
docker network create dcos_test
minidcos docker create --network dcos_test --genconf-dir genconf   ...
```
`genconf` contains the ip-detect script from https://minidcos.readthedocs.io/en/latest/dcos-docker-cli.html#using-a-custom-docker-network

This used to work before the mentioned commit. Now, the startup of exhibitor fails  => it ip-detects a node (ip on the second interface) that is not in the list of master nodes (ip on the the docker bridge).

adamtheturtle:
Great spot @brunodebus and sorry for the delay in this review. Thank you for the explanation and hunting down the commit which broke things!

CI is failing because you are outside the Mesosphere organisation and so tests which use secrets are due to fail!

PR #1784, opened 2019-05-30
Get rid of deprecated --offline flag

makkes:
This flag is deprecated and can just be removed without losing any functionality.

### DIFF
--- a/src/dcos_e2e/backends/_docker/__init__.py
+++ b/src/dcos_e2e/backends/_docker/__init__.py
@@ -572,7 +572,6 @@ class DockerCluster(ClusterManager):
         genconf_args = [
             'bash',
             str(dcos_installer),
-            '--offline',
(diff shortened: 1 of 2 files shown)


adamtheturtle:
The failed tests are easily explained.
We run an integration test in DC/OS Enterprise tests to make sure that we can.
We chose `                pytest_command=['pytest', '-vvv', '-s', '-x', 'test_tls.py'],`.

This was a quick set of tests and we captured all output.
These tests have recently grown and so they take more than 10 minutes! Travis CI has a timeout when there is no output for 10 minutes.
I will therefore:

* Change the test used
* Add log output for the test run

adamtheturtle:
Thank you @makkes - great to have less unexplainable code. I also appreciate that you created and followed this PR rather than creating an issue only!

PR #1785, opened 2019-05-30
Added VM option to customize the amount of memory given to each VM

rajcspsg:


adamtheturtle:
Thank you @rajcspsg - this looks fantastic!

PR #1788, opened 2019-05-30
Change Default Docker version on nodes to avoid issues on new Linux kernels

adamtheturtle:


src/dcos_e2e_cli/dcos_docker/commands/_docker_version.py
```diff
@@ -38,7 +39,7 @@ def docker_version_option(command: Callable[..., None],
     function = click.option(
         '--docker-version',
         type=click.Choice(sorted(_DOCKER_VERSIONS.keys())),
-        default='1.13.1',
+        default='18.06.3-ce',
```
adamtheturtle: Instead, get this from the default on the backend.

PR #1789, opened 2019-05-30
Change default memory allocation via library to be consistent with CLI

yankcrime:
The library allocates 4096MB by default to VMs instantiated via Vagrant.  This commit drops that to 2048MB so that it's consistent with the behaviour of the CLI.


adamtheturtle:
Nice, thanks!

adamtheturtle:
In future @yankcrime please create branches not forks because that will allow CI to use secrets such as a license key.

yankcrime:
> In future @yankcrime please create branches not forks because that will allow CI to use secrets such as a license key.

Noted.  I didn't have the necessary permission to push my branch to this repo, hence the fork.  I should've asked I guess!

PR #1790, opened 2019-05-30
Add --aws-instance-type option

jongiddy:
Allow the AWS instance type to be specified as an option


adamtheturtle:
Nice!

Lint issue is a yapf issue, to fix that, use `make fix-lint`.
To update the CLI tests, run `bash admin/update_cli_tests.sh`.

adamtheturtle:
Please could you also add this option to the `provision` command.

Please could you add a `CHANGELOG` entry.

adamtheturtle:
Thanks @jongiddy - this looks great. There is a big queue for CI - this wouldn't be the case here if we separated `minidcos` and `E2E` or had more parallel builders (paying Travis CI or switching to Azure Pipelines).

I cancelled the CI run because it looks good to me.
The only issue that might come up is trivial lint issues - I'll merge this and then if there is one of those I'll fix it.

PR #1798, opened 2019-06-03
Add 1.13 to the tests for legacy DC/OS versions

timaa2k:


adamtheturtle:
Thanks, let's merge to not hit the queue and then we can fix later if needed.

PR #1884, opened 2020-08-11
Add support for Flatcar OS on Docker

jongiddy:
The test for Enterprise DC/OS fails due to Flatcar using a newer version of `systemd` than other supported OS's. A fix is in progress.


jongiddy:
Tests pass with latest version of DC/OS Enterprise

